PR #11, opened 2020-05-04
Create a workflow for building documentation

hematogender:
(Also add VS Code settings folder to .gitignore)


sigma67:
Thanks! Can you elaborate a bit on how to run this .yml file? If it's a Github workflow it should be in the .github/workflows directory.

hematogender:
🤦‍♂️ Sorry, I was sleepy finishing it. You are right, let me fix it. From .github/workflows it runs automatically, exactly as ``pythonpublish.yml``. Tested it, works just fine.

P.S.: the ytmusicapi dependency can be excluded later to save a few seconds. Currently it is needed only for ``_version.py`` function, which is in sphinx's ``conf.py``. Not a lot of work, but it can certainly wait. The goal to make documentation build automatically is achieved.

PR #12, opened 2020-05-12
Implement pre-commit hooks

hematogender:
Now the following checks have to be passed before committing:

**YAPF hook check:**
Formats the staged files. If staged version differs from the correctly formatted version, the file is formatted correctly and the changes become unstaged, so the developer could review them.

**Flake8 hook check:**
Lints the staged files. If staged file had linting errors, pre-commit yields those errors and the changes become unstaged, so the developer could fix them.

If both checks were passed, commit is made.
Checks can be made explicitly with ``pre-commit run``.

I set up YAPF and Flake8, so that the code looked good enough, mostly. Also, I adjusted the  settings a little, so that the code in its current state passed the checks. Though almost all of Flake8 ignore settings are considered sane defaults, nevertheless. Line length 99 is recommended by PEP as one of options.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,10 @@
+[flake8]
+
+max-line-length = 99
+select = C,E,F,W,B,B950
+ignore = E203, E266, E731, E501, W503, F401, F403, F405
+exclude =
+    .git
+    .github
+    __pycache__
+    docs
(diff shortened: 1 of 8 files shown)


sigma67:
Hey, thanks for the contribution once again! I'll try this out the next few days and let you know my thoughts.

PR #18, opened 2020-05-16
Fix empty description in get_playlist()

### LINKED ISSUE #17: get_playlist() fails if description is blank
If I call get_playlist() on a playlist without a description I get:

```
  File "./ytmusic.py", line 100, in <module>
    main()
  File "./ytmusic.py", line 71, in main
    clear_playlist(ytmusic, playlist_id)
  File "./ytmusic.py", line 43, in clear_playlist
    playlist = ytmusic.get_playlist(id, limit=100)
  File "/usr/local/lib/python3.8/dist-packages/ytmusicapi/ytmusic.py", line 721, in get_playlist
    playlist['description'] = header['description']['runs'][0]['text']
KeyError: 'description'
```
If I fill in a description via the web UI, the call works.  After I remove the value in the web UI, the call fails again.

hematogender:
Resolves #17

### DIFF
--- a/ytmusicapi/ytmusic.py
+++ b/ytmusicapi/ytmusic.py
@@ -718,7 +718,8 @@ class YTMusic:
             header = header['header']['musicDetailHeaderRenderer']
 
         playlist['title'] = nav(header, TITLE_TEXT)
-        playlist['description'] = header['description']['runs'][0]['text']
+        if "description" in header:
+            playlist["description"] = header["description"]["runs"][0]["text"]
         run_count = len(header['subtitle']['runs'])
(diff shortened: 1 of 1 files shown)


sigma67:
Hey, I think this fix is fine for now. In the future it might be better to set the dictionary key ("description") anyway, but set it to an empty string or None. 

This makes life easier for developers building on the API.

hematogender:
I thought about that, too, but decided to mimic the YouTube API answer, which just doesn't return that key. But I agree, that the check for None/emptiness is handier. Better choose the way to handle it now, then change things later.

PR #21, opened 2020-05-18
Allow to move items within a playlist

nforro:


sigma67:
Nice, thanks for implementing this. I have thought about implementing it as well but forgot about it. 

I verified that it works and added some code to the tests.

PR #24, opened 2020-05-18
Always try to obtain setVideoId of playlist items

### LINKED ISSUE #23: It's not possible to remove items from collaborative playlists
As a collaborator, I'd like to be able to remove items from a collaborative playlist.
I'm able to do that with this [change](https://github.com/sigma67/ytmusicapi/commit/c86d4de6621ef91dc0e1098a722151c951255bff), but perhaps there is a better way?

nforro:
Collaborative playlists allow collaborators to remove items, so the items have `setVideoId` assigned, but the playlists themselves do not appear editable - they are not owned.

Make it so that attempt to obtain `setVideoId` is performed also in such cases.

Fixes #23.

### DIFF
--- a/ytmusicapi/parsers.py
+++ b/ytmusicapi/parsers.py
@@ -88,7 +88,7 @@ def parse_albums(results):
     return albums
 
 
-def parse_playlist_items(results, owned=False):
+def parse_playlist_items(results):
     songs = []
     count = 1
     for result in results:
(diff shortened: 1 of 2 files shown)


sigma67:
There was a small issue with unplayable items. I went ahead and fixed it myself.

nforro:
Thanks.

PR #36, opened 2020-07-01
Update parsers.py

joshc0:
Ran into an error where the playlist data didn't have a 'title' 
- Added a check if title exists

### DIFF
--- a/ytmusicapi/parsers.py
+++ b/ytmusicapi/parsers.py
@@ -33,8 +33,12 @@ def parse_playlists(results):
     for result in results:
         data = result['musicTwoRowItemRenderer']
         playlist = {}
-        playlist['playlistId'] = nav(data, TITLE + NAVIGATION_BROWSE_ID)[2:]
-        playlist['title'] = nav(data, TITLE_TEXT)
+        if 'title' in data:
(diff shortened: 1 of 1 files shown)


sigma67:
Can you provide an example playlist where this error occurs with the latest commit?

joshc0:
I had a playlist that didn't have a title, I think from migrating from
Google Play music

sigma67:
Could you make that playlist public and post a link here? Otherwise I can't reproduce

joshc0:
Yep will do

joshc0:
Hmm, I can't edit the permissions to make it public without giving it a
title... Happy to just delete the playlist and kill the commit

sigma67:
Yea I think this is just an edge case with the migration that they need to fix, since you can't actually create playlists with empty titles in YTMusic itself. Should probably submit to Google support. 

Nevertheless, you're welcome to submit pull requests again in the future!

joshc0:
Ah, I figured out how to do it, in case it comes up again,
https://music.youtube.com/playlist?list=PLtZPdV7nmt41QW-H01PRypm3CpExx7DUa

PR #49, opened 2020-07-20
Fix for albums without publication year

grray:
This PR fixes get_artist_albums if some albums don't have publication year. For example for https://music.youtube.com/channel/UCfUu_pLg6Uw-x_ntBWZmlwA

### DIFF
--- a/ytmusicapi/parsers.py
+++ b/ytmusicapi/parsers.py
@@ -307,9 +307,13 @@ def parse_song_album(data, index):
     }
 
 
-def get_item_text(item, index, run_index=0):
+def get_item_text(item, index, run_index=0, none_if_absent=False):
     column = get_flex_column_item(item, index)
-    return column if not column else column['text']['runs'][run_index]['text']
+    if not column: 
+        return None    
+    if none_if_absent and len(column['text']['runs']) < run_index+1:
+        return None
+    return column['text']['runs'][run_index]['text']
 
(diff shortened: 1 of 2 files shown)


sigma67:
I don't see any albums without publication year on that page, are you sure you linked the right artist? If yes, can you provide a screenshot?

grray:
It's at the bottom of album list
image

grray:
By the way, how is better for you, put these small "missing data" fixes each in separate PR, or make one big?

ytmusicapi/parsers.py
```diff
@@ -307,9 +307,13 @@ def parse_song_album(data, index):
     }
 
 
-def get_item_text(item, index, run_index=0):
+def get_item_text(item, index, run_index=0, none_if_absent=False):
     column = get_flex_column_item(item, index)
-    return column if not column else column['text']['runs'][run_index]['text']
+    if not column: 
+        return None    
+    if none_if_absent and len(column['text']['runs']) < run_index+1:
```
sigma67, line 314: Can you explain the need for ` and len(column['text']['runs']) < run_index+1` ?

grray: It's checking, how many items array has, to make sure that column['text']['runs'][run_index] exists. Perhaps there is better way to do this, python is not my main programming language :)

PR #50, opened 2020-07-21
Fix exception when there is no thumbnail and album publication year on artist page

grray:
This pull request fixes two "incomplete data" issues, each in separate commit. First is about artist, which don't have thumbnail, for example https://music.youtube.com/channel/UCDAPd3S5CBIEKXn-tvy57Lg (it doesn't have albums too, but still available in search). Second related to album without publication year, this time on artist page (https://music.youtube.com/channel/UCLZ7tlKC06ResyDmEStSrOw).

### DIFF
--- a/ytmusicapi/parsers.py
+++ b/ytmusicapi/parsers.py
@@ -91,6 +91,7 @@ class Parser:
     def parse_artist_contents(self, results: List) -> Dict:
         categories = ['albums', 'singles', 'videos', 'playlists']
         categories_local = [_('albums'), _('singles'), _('videos'), _('playlists')]
+        categories_parser = [parse_album, parse_single, parse_video, parse_playlist]
         artist = {}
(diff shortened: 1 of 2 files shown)


ytmusicapi/parsers.py
```diff
@@ -130,11 +124,20 @@ def parse_content_list(results, parse_func):
 
 
 def parse_album(result):
-    location = SUBTITLE if len(result['subtitle']['runs']) == 1 else SUBTITLE2
```
grray, line 133: `len(result['subtitle']['runs'])` is 1 for album without year, and it was detected as single with "Album" as publication year. I split this function into two to fix this.

sigma67: Good catch! I think I should include these cases in the tests or I'll keep missing them...

PR #53, opened 2020-08-01
Fix for incomplete get_library_songs request

### LINKED ISSUE #52: YTMusic responses are unreliable for get_library_songs and get_playlist
In my project I am using ytmusicapi to fetch full content of the user's library and save it in csv file. Then I can use these csv files to compare changes in my library or find songs removed from youtube etc. 

Unfortunately currently it's very unreliable.
For example:  In my library currently I have 2040 songs. To get the library songs I call the api with high limit:
```
api = YTMusic('headers_auth.json')
library_songs = api.get_library_songs(50000)
```
Everytime I send that request, the number of returned songs is different, it varies between 1800-2035 songs. 
I know that the problem is in YTM itself, because I observed the same problem on the web client and it hasn't been fixed on their side for months.  YTM should return library songs in chunks containing 25 items, but very often it's less than 25. 
In the end, on average, at least 10% of my library is missing, making my scripts kinda useless. The same problem occurs for `get_playlist` method.

czifumasa:
This is my attempt to fix #52.
When YTMusic sends incomplete response with less then per_page (25 songs), then request is retried till valid response or till max_retries number is reached.  Usually after single retry, the response is valid. I know it's not the perfect solution,  it costs performance to retry these request and complicates code a bit, but I think that more accurate results are better than fast ones. 

`resend_request_until_parsed_response_is_valid()` is generic method and can be used to validate any request. Currently I fixed only `get_library_songs call`. Let me know, what you think about it. If you approve the solution, I will prepare another fix for `get_playlist`.

### DIFF
--- a/ytmusicapi/mixins/library.py
+++ b/ytmusicapi/mixins/library.py
@@ -41,12 +41,14 @@ class LibraryMixin:
 
         return playlists
 
-    def get_library_songs(self, limit: int = 25) -> List[Dict]:
+    def get_library_songs(self, limit: int = 25, validate_responses: bool = False) -> List[Dict]:
         """
         Gets the songs in the user's library (liked videos are not included).
         To get liked songs and videos, use :py:func:`get_liked_songs`
 
         :param limit: Number of songs to retrieve
+        :param validate_responses: Flag indicating if responses from YTM should be validated and retried in case
+            when some songs are missing. Default: False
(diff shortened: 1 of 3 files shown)


czifumasa:
To give you idea, how does it work in practice, please take a look at my [export_library.log](https://github.com/sigma67/ytmusicapi/files/5010512/export_library.log)

PR #73, opened 2020-09-04
Adds support for Python's "with" statements

alexmercerind:
There is nothing mind-blowing in this little PR.

This just adds few lines to add support for following 'with syntax' in Python

```python

with YTMusic() as ytmusic:
    print(ytmusic.search('NoCopyrightSounds'))
```

Thanks a lot for your great project :smile: !


sigma67:
Merged in 78107a9

PR #90, opened 2020-10-19
Add flag with information if song from the playlist is available 

czifumasa:
ytm_missing

It happens pretty often that playlists contain unavailable/unplayable songs. In webclient these songs are greyed out, there is nothing you can do except removing them from playlist. I think API could mark those songs somehow. In my PR, I added isAvailable flag to the song object with default value `True`. My criteria for setting the flag to `False` is checking if song data has `musicItemRendererDisplayPolicy` property set to `MUSIC_ITEM_RENDERER_DISPLAY_POLICY_GREY_OUT`.  I verified the results for my library and it covered 100% of unavailable songs.

### DIFF
--- a/ytmusicapi/mixins/playlists.py
+++ b/ytmusicapi/mixins/playlists.py
@@ -47,7 +47,8 @@ class PlaylistsMixin:
                   },
                   "duration": "2:58",
                   "likeStatus": "INDIFFERENT",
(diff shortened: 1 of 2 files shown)


sigma67:
Merged in 12c465e18e601691b1ba078075d82a8e462401d4

Issue #62, opened 2020-08-20
Quota reached in ytmusicapi


xplorr:
I am creating a tool called [WinaYo](https://vanbarel.eu.org/tools/winayo/) (Winamp to Youtube Music) to convert and import WinAmp playlists (m3u8) to Youtube Music playLists (csv) using YTMusicapi. Everything is almost ready now and I started live testing with a big WinAmp library (22800 songs, 108 playlists, some playlists have over 5000 songs).
While creating the Youtube Music playlists with WinaYo, I now ran into what seems the quota limits for YTMusicapi (or Youtubeapi?): Can't create/delete playlists no more.
What are the quota-limits for YTMusicapi/Youtubeapi calls? I read somewhere that it was recently reduced from 1000000 to 40000 calls (7/9/19). Is that per day/month? Any clues? Any clues how to reset or ask for or set higher quota's?

xplorr:
Same quota problem today: api calls refused by Google Music (after calling a few times YTMusic.get_library_songs() returning about 22800 songids and calling YTMusic.get_library_playlists() returning 95 playlists . 
I will wait an hour to check again and report here.

xplorr:
update: about 8 hours later, I can again run my script with exactly the same function calls and they all return data as expected.

PR #93, opened 2020-10-28
Fix/apikey

natumbri:
I wonder if the issue arising in #62 might relate to the API key that is included in `params`? 

In his effort at implementing the YouTube music api, @DonnieWest's version grabs the `key` value dynamically.  Perhaps this would be an improvement for your version?  (Original code from @DonnieWest here: https://github.com/DonnieWest/mopidy-youtube/tree/youtubeMusic)

I've had a go at implementing it.  In `_send_request` I didn't know what `additionalParams` is for, so ... I just left it out of my implementation(!). If it is important, instead of a string that is added to the `requests.post` `url` parameter, it could be passed to `requests.post` as part of the `body` dict?

### DIFF
--- a/ytmusicapi/ytmusic.py
+++ b/ytmusicapi/ytmusic.py
@@ -14,8 +14,7 @@ from ytmusicapi.mixins.library import LibraryMixin
 from ytmusicapi.mixins.playlists import PlaylistsMixin
 from ytmusicapi.mixins.uploads import UploadsMixin
 
-params = '?alt=json&key=AIzaSyC9XL3ZjWddXya6X74dJoCTL-WEYFDNX30'
-base_url = 'https://music.youtube.com/youtubei/v1/'
+base_url = "https://music.youtube.com/youtubei/v1/"
 
 
 class YTMusic(BrowsingMixin, WatchMixin, LibraryMixin, PlaylistsMixin, UploadsMixin):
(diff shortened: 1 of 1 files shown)


sigma67:
Do you have a different API key when accessing music.youtube.com on the web? Could you post it here for comparison?

I was under the impression that all accounts used the same key. At least I had the same key when I switched to a different Google account (not brand account).

natumbri:
I had never actually checked to see what API key it was using, but it looks like the API key picked up by the code @DonnieWest wrote is the same as the API key you are using.  

I wonder if there are any advantages in grabbing the API key from a web page (as @DonnieWest's code does) vs hardcoding the API key into the code (as you have done).  I guess if the API key changes, @DonnieWest's version might continue to work whereas the hardcoding method might fail...?

DonnieWest:
It's been a bit since I last tinkered with the YouTube Music API. FWIW I've speculated that the API key is client specific (perhaps even client-version specific) but I have no evidence to prove it

Either way, a hybrid approach might make sense: hard code the token, retrieve a new one if it's invalid

* Retrieving it every time might mask API updates that one should be aware of (even if there are no breaking changes)
* Retrieving it every time leads to unnecessary API calls, bandwidth and increased start up time
* Strictly hard coding it might cause unnecessary breakage where there's an update but everything remains compatible

Take all of this with a grain of salt. As mentioned, I've not touched YouTube Music in a while and so this might be based on outdated info

sigma67:
At this point in time, I have a couple issues with this PR:

- There is no proof that it solves any problem. The current API key seems to work for all users. Suggestions that it may not work in the future are speculation at best - we can come back to the PR if it does turn out to be the case
- The PR leads to failing tests, at least on my machine. It also introduces at least one additional request for each instantiation of YTMusic, which would introduce latency
- The commits change lines of code that have nothing to do with the PR's purpose, making it difficult to comprehend what was actually changed

In summary, currently I don't see anything positive for the project that could come from accepting this PR. Barring new information or developments, that is.

natumbri:
Yes, that makes sense. I'll close this now; if it becomes relevant in the future, we'll know where to find it.

Issue #94, opened 2020-10-28
More than 20 results from a search?


natumbri:
Any interest in returning more than 20 results from a search?

I've had a go, and it seems to work: https://github.com/natumbri/ytmusicapi/tree/fix/more-results

Let me know if you'd like a pull request.  I haven't tested it with authorisation.

sigma67:
Yes, this is one thing I didn't get around to implementing. Great that you've gone ahead with it. 

I looked at your code and it looks good, I have just a few suggestions for improvement. Please submit a PR and I will comment there

natumbri:
Great - I'll try to make a PR. 

I'm pretty hopeless at them, particularly keeping the automatic linting tools under control, and I haven't been able to make the unittests work at all. But I'll have a crack and we'll see how it goes...

PR #95, opened 2020-10-28
add limit to search()

natumbri:
As discussed in #94, here is a pull request that adds a `maxResults` parameter to `search()`.

### DIFF
--- a/ytmusicapi/mixins/browsing.py
+++ b/ytmusicapi/mixins/browsing.py
@@ -4,19 +4,22 @@ import codecs
 from urllib.parse import parse_qs
 from typing import List, Dict
 from ytmusicapi.helpers import *
+from ytmusicapi.parsers.browsing import parse_continuation_content_list
 from ytmusicapi.parsers.playlists import *
 
 
 class BrowsingMixin:
-    def search(self, query: str, filter: str = None) -> List[Dict]:
+    def search(self, query: str, filter: str = None, limit: int = 20) -> List[Dict]:
         """
         Search YouTube music
-        Returns up to 20 results within the provided category.
+        Returns results within the provided category.
 
         :param query: Query string, i.e. 'Oasis Wonderwall'
         :param filter: Filter for item types. Allowed values:
           'songs', 'videos', 'albums', 'artists', 'playlists', 'uploads'.
           Default: Default search, including all types of items.
+        :param limit: Number of search results to return
+          Default: 20
         :return: List of results depending on filter.
           resultType specifies the type of item (important for default search).
           albums, artists and playlists additionally contain a browseId, corresponding to
@@ -106,10 +109,21 @@ class BrowsingMixin:
                         search_result = self.parser.parse_search_result(data, type)
                         search_results.append(search_result)
 
+                if 'continuations' in res['musicShelfRenderer']:
+                    request_func = lambda additionalParams: self._send_request(
+                        endpoint, body, additionalParams)
+
+                    parse_func = lambda contents: parse_continuation_content_list(
+                        contents, self.parser.parse_search_result)
+
+                    search_results.extend(
+                        get_continuations(res['musicShelfRenderer'], 'musicShelfContinuation',
+                                          limit - len(search_results), request_func, parse_func))
+
         except Exception as e:
             print(str(e))
 
-        return search_results
+        return search_results[:limit]
 
     def get_artist(self, channelId: str) -> Dict:
         """
@@ -436,7 +450,9 @@ class BrowsingMixin:
                 # in case the song is unavailable, there is no videoId
                 track['videoId'] = item['payload']['musicTrack']['videoId'] if 'videoId' in item[
                     'payload']['musicTrack'] else None
-                track['lengthMs'] = item['payload']['musicTrack']['lengthMs']
+                # very occasionally lengthMs is not returned
+                track['lengthMs'] = item['payload']['musicTrack'][
+                    'lengthMs'] if 'lengthMs' in item['payload']['musicTrack'] else None
                 track['likeStatus'] = likes[item['entityKey']]
                 album['tracks'].append(track)
 
--- a/ytmusicapi/parsers/browsing.py
+++ b/ytmusicapi/parsers/browsing.py
@@ -29,7 +29,46 @@ class Parser:
             search_result['title'] = get_item_text(data, 0)
 
         if resultType in ['artist', 'album', 'playlist']:
-            search_result['browseId'] = nav(data, NAVIGATION_BROWSE_ID)
+            browse_id = nav(data, NAVIGATION_BROWSE_ID, True)
+            if not browse_id:  # song result
+                search_result['videoId'] = nav(
+                    data, PLAY_BUTTON)['playNavigationEndpoint']['watchEndpoint']['videoId']
+                search_result['playlistId'] = nav(
+                    data, PLAY_BUTTON)['playNavigationEndpoint']['watchEndpoint']['playlistId']
+                flex_items = [
+                    nav(get_flex_column_item(data, i), ['text', 'runs'], True) for i in range(4)
+                ]
+                if flex_items[0]:
+                    search_result['title'] = {flex_items[0][0]['text']}
+                if flex_items[1]:
+                    search_result['artists'] = [{
+                        'name': flex_items[1][0]['text'],
+                        'id': nav(flex_items[1][0], NAVIGATION_BROWSE_ID)
+                    }]
+                if flex_items[2]:
+                    search_result['album'] = {
+                        'name': flex_items[2][0]['text'],
+                        'id': nav(flex_items[2][0], NAVIGATION_BROWSE_ID)
+                    }
+                if flex_items[3]:
+                    search_result['duration'] = flex_items[3][0]['text']
+                search_result['resultType'] = 'song'
+            else:  # artist or album result
+                search_result['browseId'] = browse_id
+                if 'artist' in search_result['browseId']:
+                    search_result['resultType'] = 'artist'
+                else:
+                    flex_items2 = [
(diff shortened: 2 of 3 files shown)


ytmusicapi/mixins/browsing.py
```diff
@@ -75,41 +98,60 @@ def search(self, query: str, filter: str = None) -> List[Dict]:
...
-            for res in results:
-                if 'musicShelfRenderer' in res:
-                    results = res['musicShelfRenderer']['contents']
+        try:
+            results = None
+
+            if "continuationContents" in response:
+                results = response["continuationContents"]
+            elif "contents" in response:
+                if "tabbedSearchResultsRenderer" in response["contents"]:
```
natumbri, line 117: Is this part, dealing with `"tabbedSearchResultsRenderer"` correct?  I've never seen a response that includes `"tabbedSearchResultsRenderer"` so I have no idea if it works.

ytmusicapi/ytmusic.py
```diff
@@ -127,9 +132,9 @@ def setup(cls, filepath: str = None, headers_raw: str = None) -> Dict:
...
         """
         return setup(filepath, headers_raw)
-    
+
     def __enter__(self):
         return self
-    
-    def __exit__(self, execType = None, execValue = None, trackback = None):
+
+    def __exit__(self, execType=None, execValue=None, trackback=None):
```
natumbri, line 139: `yapf` did this, I think, and I couldn't work out how to undo it

ytmusicapi/ytmusic.py
```diff
@@ -127,9 +132,9 @@ def setup(cls, filepath: str = None, headers_raw: str = None) -> Dict:
         :return: configuration headers string
         """
         return setup(filepath, headers_raw)
-    
+
```
natumbri, line 135: `yapf` did this, I think, and I couldn't work out how to undo it

ytmusicapi/mixins/browsing.py
```diff
@@ -8,15 +8,17 @@
 
 
 class BrowsingMixin:
-    def search(self, query: str, filter: str = None) -> List[Dict]:
+    def search(self, query: str, filter: str = None, maxResults: int = 20) -> List[Dict]:
```
sigma67, line 11: This parameter should be named limit to be consistent with other function calls

ytmusicapi/ytmusic.py
```diff
@@ -100,12 +100,17 @@ def __init__(self,
                     "The provided credentials are invalid. Reason given by the server: "
                     + response['error']['status'])
 
-    def _send_request(self, endpoint: str, body: Dict, additionalParams: str = "") -> Dict:
+    def _send_request(self,
```
sigma67, line 103: There should be no need to modify this file as part of this PR. I recommend to use the battle-tested get_continuations function here: https://github.com/sigma67/ytmusicapi/blob/ca8a73046bb43441d8e40a84569e0383bcb305f6/ytmusicapi/parsers/utils.py#L50-L59

For usage, refer for example to https://github.com/sigma67/ytmusicapi/blob/5a1032983d1a665019944957aeb5eb4886e799b0/ytmusicapi/mixins/library.py#L8

It has all the continuation logic you need. This avoids duplicate code in the project, which currently there is quite a bit of in this PR. I have marked it where applicable.

ytmusicapi/mixins/browsing.py
```diff
@@ -75,41 +98,60 @@ def search(self, query: str, filter: str = None) -> List[Dict]:
...
-                    results = res['musicShelfRenderer']['contents']
+        try:
+            results = None
+
+            if "continuationContents" in response:
+                results = response["continuationContents"]
+            elif "contents" in response:
+                if "tabbedSearchResultsRenderer" in response["contents"]:
+                    results = response["contents"]["tabbedSearchResultsRenderer"]["tabs"][int(
+                        filter ==  # noqa: W504
```
sigma67, line 119: This looks really odd. Try to put the comparison in one line. There should be no need for noqa

ytmusicapi/mixins/browsing.py
```diff
@@ -75,41 +98,60 @@ def search(self, query: str, filter: str = None) -> List[Dict]:
...
+                if "continuations" in musicShelf:
+                    nextToken = musicShelf["continuations"][0]["nextContinuationData"][
+                        "continuation"]
+            else:
+                for content in results:
+                    musicShelf = content["musicShelfRenderer"]
+                    items.append(musicShelf["contents"])
+                    if musicShelf["continuations"]:
+                        nextToken = musicShelf["continuations"][0]["nextContinuationData"][
+                            "continuation"]
```
sigma67, line 141: This part will probably not be needed when you use get_continuations

ytmusicapi/ytmusic.py
sigma67: It's fine, don't worry about it

ytmusicapi/ytmusic.py
sigma67: dw

ytmusicapi/mixins/browsing.py
sigma67: It happens during upload search. Please upload at least a test song on your account so you can verify that the upload search works

ytmusicapi/mixins/browsing.py
```diff
@@ -44,10 +46,31 @@ def search(self, query: str, filter: str = None) -> List[Dict]:
...
+        continuationToken = None
+        try:
+            while len(search_results) < maxResults:
+                result = self._base_search(q=query,
+                                           filter=filter,
+                                           continuationToken=continuationToken)
+                search_results.extend(result["search_results"])
+                continuationToken = result["nextPageToken"]
+                if continuationToken is None:
+                    break
```
sigma67, line 59: This logic is already in get_continuations

sigma67 (changes requested):
I think you could simplify a lot of things by using get_continuations. Additionally, please make sure that test_search and test_search_uploads work. For test_search_uploads you might need to modify the search query (locally, dont commit) so that it works for your account.

ytmusicapi/mixins/browsing.py
natumbri: Thanks - I thought it might be something like that.  I'll try, but I haven't had any success getting the tests to work.  No doubt some misunderstanding/misconfiguration at my end.

ytmusicapi/mixins/browsing.py
natumbri: Will do.

ytmusicapi/ytmusic.py
natumbri: Great - I'll give this a go.  Thanks.

ytmusicapi/mixins/browsing.py
natumbri: Yes, it really does look strange - the automatic formatting tool went crazy and I couldn't work out how to get it to pass without the noqa.  I'll try again.

ytmusicapi/mixins/browsing.py
natumbri: Noted.

ytmusicapi/mixins/browsing.py
natumbri: Noted.

sigma67:
If you have any questions regarding the test setup or otherwise we can chat on Gitter (https://gitter.im/sigma67/ytmusicapi). That should be easier and quicker. 

I know the tests can be frustrating to set up as they heavily depend on what music you have on your account. The documentation could probably also be improved.

natumbri:
Thanks - will do. I've really struggled to get it to work using your existing functions. I feel like you've created a bunch of tools to do exactly what I need to do, but I don't know how to use them!

ytmusicapi/mixins/browsing.py
```diff
@@ -106,10 +109,21 @@ def search(self, query: str, filter: str = None) -> List[Dict]:
...
+
+                    search_results.extend(
+                        get_continuations(res['musicShelfRenderer'], 'musicShelfContinuation',
+                                          limit - len(search_results), request_func, parse_func))
+
         except Exception as e:
             print(str(e))
 
-        return search_results
+        return search_results[:limit]
```
sigma67, line 126: The convention in ytmusicapi is to return all retrieved results regardless of the limit, so no need to subset here

ytmusicapi/parsers/browsing.py
```diff
@@ -29,15 +29,56 @@ def parse_search_result(self, data, resultType=None):
...
+                        nav(get_flex_column_item(data, i), ['text', 'runs'], True)
+                        for i in range(4)
+                    ]
+                    if flex_items2[0]:
+                        search_result['title'] = flex_items2[0][0]['text']
+                    if flex_items2[2]:
+                        search_result['artist'] = {'name': flex_items2[2][0]['text'], 'id': None}
+                    if flex_items2[3]:
+                        search_result['releaseDate'] = flex_items2[3][0]['text']
+                    search_result['resultType'] = 'album'
```
sigma67, line 71: What is the reason for adding this code? Seems it's only copied from further down?

IMO there is no need to change the parser at all, as the continuation items should have an identical format. Feel free to correct me if I'm missing something here

sigma67 (commented):
Thanks for the changes! Looks almost ready. Just the changes in the browsing parser look a little odd to me, maybe you can explain.

ytmusicapi/parsers/browsing.py
natumbri: Apologies I never got back to you on this - real life got in the way a bit! Did you end up needing to change the parser as I had thought? 

Thanks for implementing, in any event - I might switch mopidy-youtube over to your implementation rather than the one I'm currently - yours seems very robust. Or possibly make a proper mopidy-ytmusic, with which includes vthe authenticated functions, too.

Anyway, thanks again!

PR #98, opened 2020-11-01
Fixes #46: Add X-Goog-AuthUser to required headers

### LINKED ISSUE #46: Match 'X-Goog-AuthUser' during setup.
Thanks for writing this! I was playing around with it and noticed that during setup it does not copy the number from the request header X-Goog-AuthUser. 

In the pasted headers from firefox contained:

```
X-Goog-AuthUser: 2
```

But the generated header_auth.json was using:

```
"X-Goog-AuthUser": "0"
``` 

This was causing my liked songs, playlists, etc to be pulled from different google account that I don't use with youtube music. 

I have a few google accounts. I'm not sure how google handles all that. This particular account is not a brand account, and I didn't notice this in the docs so thought I'd let you know. 

I noticed the AuthUser number matches this url when selecting the correct account: https://myaccount.google.com/u/2/?pli=1 (the 2 changes to 0 on my default account). 

I changed the value X-Goog-AuthUser to match the firefox request headers and it worked perfectly. 

Thanks again! Let me know if you need more info.

gaganpreet:


sigma67:
Merged in 7974715

PR #103, opened 2020-11-11
Fix #102: Add sort order option to library functions

### LINKED ISSUE #102: Provide sort order option to get_library_upload_songs()
As mentioned in https://github.com/sigma67/ytmusicapi/issues/63#issuecomment-678684815, we need sort order options on the uploaded songs list (especially,  `most recently added`) .
YouTube Music has already provided sort options, so we can implement it in ytmusicapi.
sort_order_ui

Here is my implementation.
https://github.com/ide-an/ytmusicapi/commit/4a853c6555621bc7c2a323bfff4b98d2a2823f3b

ide-an:
fixes #102 
This pullrequest adds `order` parameter to functions bellow:
 * `get_library_songs`
 * `get_library_albums`
 * `get_library_artists`
 * `get_library_subscriptions`
 * `get_library_upload_songs`
 * `get_library_upload_albums`
 * `get_library_upload_artists`

`order`  parameter accepts `"a_to_z"`, `"z_to_a"` or `"recently_added"`, and the function  returns response in the specified order.

----

I once implemented for `get_library_playlists` too, but I faced a problem and gave up.
When you call `get_library_playlists` with `order` parameter, youtube music API returns *multiple "New playlist" buttons* and causes parse error.

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -215,6 +215,12 @@ class TestYTMusic(unittest.TestCase):
     def test_get_library_upload_songs(self):
         results = youtube_auth.get_library_upload_songs(100)
         self.assertGreater(len(results), 25)
+        results = youtube_auth.get_library_upload_songs(100, 'a_to_z')
+        self.assertGreater(len(results), 25)
+        results = youtube_auth.get_library_upload_songs(100, 'z_to_a')
+        self.assertGreater(len(results), 25)
+        results = youtube_auth.get_library_upload_songs(100, 'recently_added')
+        self.assertGreater(len(results), 25)
 
     @unittest.skip("Must not have any uploaded songs to pass")
     def test_get_library_upload_songs_empty(self):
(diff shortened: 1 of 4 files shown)


ide-an:
> When you call get_library_playlists with order parameter, youtube music API returns multiple "New playlist" buttons and causes parse error.

You can observe this problem on YouTube Music UI.

how to reproduce:
 1. Open https://music.youtube.com/library/playlists
 2. Select sort option dropdown and change sort order
 3. If you have many playlists, multiple "New playlist" buttons appear.

playlist_button_problem

sigma67:
Awesome, this is very thorough. Thanks for implementing it for all the functions and even adding tests.

The "New playlist" button issue is quite strange, I doubt that is intended behavior. I suggest we wait a bit to see if this issue gets fixed by YouTube, else we could probably implement a workaround for that specific case.

sigma67:
Merged in a0506ebf.

Small note: Might remove some of the parameter tests later since they increased the total runtime of the tests considerably (2x-3x). Don't think we need to test every possible parameter combination in the long term

PR #108, opened 2020-11-24
Retreive streaming data for 256 kbps AAC

corrmaan:
Adding the parameters `c=WEB_REMIX` and `cver=0.1` to the `get_video_info` request in `get_song()` returns a `streamingData` `adaptiveFormat` of `itag` `141`, `audioQuality` of `AUDIO_QUALITY_HIGH`, a 256 kbps AAC m4a for authenticated requests from YouTube Music Premium subscribers.

Unfortunately, these parameters result in `player_response['videoDetails']['shortDescription']` not being returned, the information in which contains useful song metadata.

`player_response['microformat']['microformatDataRenderer']['description']` is returned, but it seems to be a truncated version of `shortDescription`, so the metadata parsing in `get_song()` does not work on that response.

Therefore, a second `get_video_info` request is made with the updated parameters and the metadata in `song_meta` is copied in to the new `player_response['videoDetails']`, which is returned from `get_song()` instead of `song_meta`.

### DIFF
--- a/docs/source/reference.rst
+++ b/docs/source/reference.rst
@@ -27,6 +27,7 @@ Browsing
 .. automethod:: YTMusic.get_user_playlists
 .. automethod:: YTMusic.get_album
 .. automethod:: YTMusic.get_song
+.. automethod:: YTMusic.get_streaming_data
 
 Watch
 --------
--- a/tests/test.py
+++ b/tests/test.py
@@ -84,8 +84,8 @@ class TestYTMusic(unittest.TestCase):
         self.assertGreaterEqual(len(song), 16)
 
     def test_get_streaming_data(self):
-        song = youtube.get_streaming_data("NRUVdGLf2GI")
-        self.assertEqual(len(song), 3)
+        streaming_data = youtube.get_streaming_data("NRUVdGLf2GI")
+        self.assertEqual(len(streaming_data), 3)
 
     ###############
     # WATCH
--- a/tests/test.py
+++ b/tests/test.py
@@ -81,7 +81,11 @@ class TestYTMusic(unittest.TestCase):
 
     def test_get_song(self):
         song = youtube.get_song("ZrOKjDZOtkA")
-        self.assertGreaterEqual(len(song), 17)
+        self.assertGreaterEqual(len(song), 16)
+
+    def test_get_streaming_data(self):
+        song = youtube.get_streaming_data("NRUVdGLf2GI")
+        self.assertEqual(len(song), 3)
 
     ###############
     # WATCH
--- a/ytmusicapi/mixins/browsing.py
+++ b/ytmusicapi/mixins/browsing.py
@@ -526,7 +526,7 @@ class BrowsingMixin:
                 pass
         return song_meta
 
-    def get_streamingData(self, videoId: str) -> Dict:
+    def get_streaming_data(self, videoId: str) -> Dict:
         """
         Returns the streaming data for a song or video.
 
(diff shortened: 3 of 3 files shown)


sigma67:
Great suggestion. 

One minor comment: This introduces an additional request that might not be needed for users interested in metadata. I think it might be better to create a separate function for this named `get_stream_url`, inspired by gmusicapi's equivalent. What do you think?

corrmaan:
Yes I think that's a good way to go about it. I'll work on modifying the PR.

corrmaan:
Perhaps just a `get_streamingData` function would be good for now? Since url parsing is a fair bit of processing in itself (see https://github.com/ytdl-org/youtube-dl/blob/master/youtube_dl/extractor/youtube.py#L1886).

sigma67:
Yeah, although I'd call it `get_streaming_data` to follow Python naming conventions. The idea is to just return what's returned by the YouTube API call. 

I don't want to duplicate any youtube-dl logic. Initially I wanted to avoid interacting with the YouTube-specific endpoints at all, but song metadata is quite important to a lot of users.

corrmaan:
Yeah YouTube Music seems like more of just a new front-end for YouTube, oh well. Changes made for your review, thanks for your comments.

corrmaan:
Sorry I forgot about this, but if you'd like I can also add commits for the documentation and tests.

sigma67:
Regarding the code and completeness of the PR, everything looks good. I'm just having issues running the test successfully. The streamingData key is missing in my response.

At first I thought this was due to the non-authenticated instance `youtube` being used instead of `youtube_auth`, but even changing that didn't fix it. Does it work for you?

corrmaan:
Interesting, not sure why it would be missing for you. Here's a minimal working example below. I tested it with the defaults in `ytmusicapi/headers_auth.json.example` with my own cookie too.

```
import unittest
import configparser
import sys
sys.path.insert(0, '..')
from ytmusicapi.ytmusic import YTMusic  # noqa: E402

config = configparser.RawConfigParser()
config.read('./test.cfg', 'utf-8')

youtube = YTMusic()
youtube_auth = YTMusic(config['auth']['headers_file'])


class TestYTMusic(unittest.TestCase):
    def test_init(self):
        self.assertRaises(Exception, YTMusic, "{}")

    def test_setup(self):
        YTMusic.setup(config['auth']['headers_file'], config['auth']['headers_raw'])

    ###############
    # BROWSING
    ###############

    def test_get_song(self):
        song = youtube.get_song("ZrOKjDZOtkA")
        self.assertGreaterEqual(len(song), 16)

    def test_get_song_auth(self):
        song = youtube_auth.get_song("ZrOKjDZOtkA")
        self.assertGreaterEqual(len(song), 16)

    def test_get_streaming_data(self):
        streaming_data = youtube.get_streaming_data("NRUVdGLf2GI")
        self.assertEqual(len(streaming_data), 3)

    def test_get_streaming_data_auth(self):
        streaming_data = youtube_auth.get_streaming_data("NRUVdGLf2GI")
        self.assertEqual(len(streaming_data), 3)


if __name__ == '__main__':
    unittest.main()
```

corrmaan:
Maybe those new URL parameters only work with a cookie associated with a premium account?

sigma67:
Ah, I've found the issue by checking the videoId on YouTube itself. The video is unavailable in my country. Which is actually a good thing, because it helped us find this issue before it was merged. I'll implement some error handling for this edge case and merge afterward.

Thanks for your contribution! :)

corrmaan:
Ah, glad we found that! You're welcome, and thanks for the discussion.

corrmaan:
Would you like me to change the videoId to the same one as in the `get_song()` test?

sigma67:
Did that just before you commented. Merged in 028b678.

Issue #127, opened 2020-12-17
KeyError from "search()", missing navigationEndpoint in album result


wsidl:
## Description
Running a search for track titles also returns album and artist results. "KeyError" is raised before the search function finishes.

## Expected behaviour
```python
results = yt.search("That's My Jam Skindred That's My Jam")
# parse results
```

## Current behaviour
```python
results = yt.search("That's My Jam Skindred That's My Jam")
Traceback (most recent call last):
  File "/usr/lib/python3.8/site-packages/ytmusicapi/parsers/utils.py", line 167, in nav
    raise err
  File "/usr/lib/python3.8/site-packages/ytmusicapi/parsers/utils.py", line 161, in nav
    root = root[k]
KeyError: 'navigationEndpoint'
```

Stack Trace:
```
  File "/scratch.py", line 110, in find
    results = YT.search(search_term)
  File "/usr/lib/python3.8/site-packages/ytmusicapi/mixins/browsing.py", line 119, in search
    search_results.extend(self.parser.parse_search_results(results, type))
  File "/usr/lib/python3.8/site-packages/ytmusicapi/helpers.py", line 78, in _impl
    return method(self, *method_args, **method_kwargs)
  File "/usr/lib/python3.8/site-packages/ytmusicapi/parsers/browsing.py", line 36, in parse_search_results
    search_result['browseId'] = nav(data, NAVIGATION_BROWSE_ID)
  File "/usr/lib/python3.8/site-packages/ytmusicapi/parsers/utils.py", line 167, in nav
    raise err
  File "/usr/lib/python3.8/site-packages/ytmusicapi/parsers/utils.py", line 161, in nav
    root = root[k]
KeyError: 'navigationEndpoint'
```

## Additional Points
It does successfully parse many results, but gets stuck on an album entry.

```
nav(root, items, none_if_absent)
none_if_absent = False
items = ['navigationEndpoint', 'browseEndpoint', 'browseId']
root = {
	"trackingParams": "CLcBENjyAxgAIhMInoX5jfPV7QIV1YDECh09kgSk",
	"thumbnail": {
		"musicThumbnailRenderer": {
			"thumbnail": {
				"thumbnails": [
[...]

PR #128, opened 2020-12-17
Skip entry if item is private or uploaded

wsidl:
Possible solution for identifying invalid/uploaded entries and skip from results.


wsidl:
Solves #127

sigma67:
Merged in 95d69fe6

PR #153, opened 2021-01-17
Add radio/Shuffle ID to search results

PhracturedBlue:
I wanted to be able to create streaming radio channels from search results.
This patch extracts the Shuffle and Radio playlist ids from the search results.

yapf seems to have changed an unrelated line for some reason, I can manually revert it if you like

I was unable to get all the tests to pass (even without my change), but the search tests all seem to work ok

### DIFF
--- a/ytmusicapi/parsers/browsing.py
+++ b/ytmusicapi/parsers/browsing.py
@@ -71,7 +71,8 @@ class Parser:
                         search_result['feedbackTokens'] = parse_song_menu_tokens(toggle_menu)
 
             elif resultType in ['video']:
-                search_result['views'] = get_item_text(data, 1, 2 + last_artist_index).split(' ')[0]
+                search_result['views'] = get_item_text(data, 1,
+                                                       2 + last_artist_index).split(' ')[0]
                 search_result['duration'] = get_item_text(data, 1, 4 + last_artist_index, True)
 
             elif resultType in ['upload']:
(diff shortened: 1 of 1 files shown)


sigma67:
I really like this idea, I think we could add this code across the board. Most endpoints actually have these menu entries and it shouldn't break any existing functionality.

sigma67:
As you suggested in a comment, we should definitely use the iconType to determine the correct menu item. The text which is used now is sensitive to localization ("hl" setting in headers). Feel free to update the PR accordingly.

PhracturedBlue:
I have updated the pattern to use the icon name which is hopefully more robust to locales.

sigma67:
Thanks for the update. I experimented a bit more and found that getting the shuffle id is pointless. It's always the same as the playlist id. Instead you should call `get_watch_playlist_shuffle` with the playlist id, that's what happens behind the scenes when you press shuffle. Difference to normal play is that it only adds the hard coded param `wAEB8gECGAE%3D`, which is always the same for any type of shuffle.

sigma67:
I think we should go ahead and implement shuffle/radio ID for artist results only. 

For all others, I will create a new top level docs page that includes info on intricacies of the YTMusic API, which would include the composition of radio IDs. Should be very helpful to new developers.

PhracturedBlue:
I was primarily just concerned with artist, so that makes sense.

PR #174, opened 2021-03-12
Switch youtube /browse endpoint with /search

YuviGold:


sigma67:
Why? For search to show up you would have to perform a search first, browse requests show up when you just go to the homepage or library.

YuviGold:
Sorry for the lack of description.

I thought that the endpoint might have changed and I couldn't find browse API.
When I am entering Youtube Music homepage I don't see any "browse" request.

So the next thing, I figured I had to make a manual search. That's written in the docs as well:
```
- Find an authenticated POST request. The simplest way is to filter by ``/browse`` using the search bar
```
When using the search bar - the "search" API is being requested.
For the "browse" API to be used, I had to scroll over the homepage until the cache wasn't enough to be shown.

So either way, in case the documentation keeps guiding to use the search bar - the use should look for the "search" API :)

sigma67:
Thanks for the clarification. It seems there is a misunderstanding - in the docs I am talking about the search bar within the developer tools, where you can filter applicable requests. Not the search bar in YouTube Music. Perhaps the docs should be updated a bit to clarify that

YuviGold:
> Thanks for the clarification. It seems there is a misunderstanding - in the docs I am talking about the search bar within the developer tools, where you can filter applicable requests. Not the search bar in YouTube Music. Perhaps the docs should be updated a bit to clarify that

I see. This is quite misleading.

I just noticed that the browser resolution affects whether the "browse" API is shown immediately when the homepage is being rendered.
I'm not sure from which element it is being asked. But depends on the resolution (and probably that's what happened for me at the first time) - the "browse" API might not appear when you land on the homepage and you will have to do another manual operation (scroll) for it to show up.

Issue #176, opened 2021-03-18
After adding songs to a playlist: return the new setVideoId for each song


MattC8915:
The setVideoId is helpful information to have.

in `add_playlist_items` instead of just returning `response['status']` also include the `playlistEditResults`

I would be happy to submit a pull request if you would accept it. But this would break the pattern of only returning STATUS_SUCCEEDED for successful requests...is that acceptable?

sigma67:
Good point, that could be useful. Feel free to submit a PR. Please also test with source_playlist parameter and multiple videoIds, as well as the dedupe option (duplicates parameter).

sigma67:
Merged in #180

PR #180, opened 2021-03-21
After adding songs to a playlist: return the new setVideoId for each song

MattC8915:
For this enhancement: https://github.com/sigma67/ytmusicapi/issues/176

### DIFF
--- a/ytmusicapi/mixins/playlists.py
+++ b/ytmusicapi/mixins/playlists.py
@@ -222,7 +222,7 @@ class PlaylistsMixin:
         :param videoIds: List of Video ids
         :param source_playlist: Playlist id of a playlist to add to the current playlist (no duplicate check)
         :param duplicates: If True, duplicates will be added. If False, an error will be returned if there are duplicates (no items are added to the playlist)
-        :return: Status String and a dict that maps the videoIds to their new setVideoIds or full response
+        :return: Status String and a dict containing the new setVideoId for each videoId or full response
         """
         self._check_auth()
         body = {'playlistId': playlistId, 'actions': []}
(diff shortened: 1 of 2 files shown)


ytmusicapi/mixins/playlists.py
```diff
@@ -270,3 +274,17 @@ def remove_playlist_items(self, playlistId: str, videos: List[Dict]) -> Union[st
         endpoint = 'browse/edit_playlist'
         response = self._send_request(endpoint, body)
         return response['status'] if 'status' in response else response
+
```
sigma67, line 277: Please don't create a utility function for it. Two reasons:

- the functionality is not used anywhere else
- it would be exposed to the library caller, which is not needed

Instead just put the code right into `add_playlist_items`.

ytmusicapi/mixins/playlists.py
```diff
@@ -232,6 +232,10 @@ def add_playlist_items(self,
                 action['dedupeOption'] = 'DEDUPE_OPTION_SKIP'
             body['actions'].append(action)
 
+        # add an empty ACTION_ADD_VIDEO because otherwise YTM doesn't return the dict that maps videoIds to their new setVideoIds
+        if not videoIds:
```
sigma67, line 236: I assume this is needed only if a source playlist is used?

sigma67 (changes requested):
Thanks for the PR! Just two small comments :)

ytmusicapi/mixins/playlists.py
MattC8915: yep

PR #181, opened 2021-03-22
Feature: Related Artists in get_artist and Exception for Non-Music Channels

mikefreemanwd:
In the `BrowserMixin` class from the mixins/browsing.py file there is a method called `get_artist` that returns artist information. 

I made 2 modifications to `get_artist`:
1. Return the related artists that appear at the bottom of the page under the "Fans might also like" section. These related artists are returned as a list of dictionaries that contain the artist name, the artist YouTube Music Channel ID, and the text that shows the approximate number of subscribers
2. You can technically visit any YouTube channel in the YouTube music front-end. That means you can visit non-music channel IDs that will not return any YouTube music content (example: [Google Developers](https://music.youtube.com/channel/UC_x5XG1OV2P6uZZ5FSM9Ttw)). I created an exception called `YouTubeChannelIDIsNotArtistException` that is thrown when a non-music channel ID is provided to the `get_artist` function. 

I updated the documentation for the `get_artist` function, added a unit test for each modification, and made sure that the pre-commit linters pass.

This can be called with:
```python
ytmusic = YTMusic()
youtube_channel_id = "..."
return ytmusic.get_artist(youtube_channel_id)
```

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -2,7 +2,7 @@ import unittest
 import configparser
 import sys
 sys.path.insert(0, '..')
-from ytmusicapi.ytmusic import YTMusic, YouTubeChannelIDIsNotArtistException  # noqa: E402
+from ytmusicapi.ytmusic import YTMusic  # noqa: E402
 
 config = configparser.RawConfigParser()
 config.read('./test.cfg', 'utf-8')
@@ -55,16 +55,11 @@ class TestYTMusic(unittest.TestCase):
         # test correctness of related artists
         related = results['related']['results']
 
-        # all should have 3 keys per item
         self.assertEqual(
-            len([x for x in related if set(x.keys()) == {"browseId", "subscribers", "title"}]),
-            len(related))
-
-        # test names of some related artists
-        related_artist_names = {x['title'] for x in related}
-        self.assertTrue(
-            {"The Verve", "Liam Gallagher", "Noel Gallagher", "Beady Eye",
-             "The Stone Roses"}.issubset(related_artist_names))
+            len([
+                x for x in related
+                if set(x.keys()) == {"browseId", "subscribers", "title", "thumbnails"}
+            ]), len(related))
 
         results = youtube.get_artist("UCLZ7tlKC06ResyDmEStSrOw")  # no album year
         self.assertGreaterEqual(len(results), 11)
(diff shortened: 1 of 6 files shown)


ytmusicapi/parsers/browsing.py
```diff
@@ -198,3 +198,12 @@ def parse_playlist(data):
     if len(data['subtitle']['runs']) == 3:
         playlist['count'] = nav(data, SUBTITLE2).split(' ')[0]
     return playlist
+
+
+def parse_related_artist(data):
+    item = data['musicTwoRowItemRenderer']
+    return {
```
sigma67, line 205: Should also include the thumbnails here, see for example `parse_artists`

tests/test.py
```diff
@@ -50,13 +50,34 @@ def test_search_uploads(self):
...
+
+        # test correctness of related artists
+        related = results['related']['results']
+
+        # all should have 3 keys per item
+        self.assertEqual(
+            len([x for x in related if set(x.keys()) == {"browseId", "subscribers", "title"}]),
+            len(related))
+
+        # test names of some related artists
```
sigma67, line 63: I'd advise against testing artist names specifically since the list may change for unrelated reasons (different recommendations), causing the test to break

sigma67 (changes requested):
Thanks again for the contribution, good idea on adding the related artists as well. I left just a few comments.

ytmusicapi/helpers.py
```diff
@@ -83,3 +83,11 @@ def _impl(self, *method_args, **method_kwargs):
         return method(self, *method_args, **method_kwargs)
 
     return _impl
+
+
+class YouTubeChannelIDIsNotArtistException(Exception):
```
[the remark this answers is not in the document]
mikefreemanwd: This is more of a stylistic choice to make the error a bit more specific. 

Thinking about it again, I think an exception like ValueError with a message that explains what happened would be better given that the library does not return any other named exceptions.

sigma67:
There's still an issue with the way you're finding the index of the related artists. Numeric index doesn't work here because different artists have different segments available (see i.e. UC3xSIzoyzI_8XouBBN2mVeA vs. UCmMUZbaYdNH0bEd1PAlAqsA). Instead you need to check the header text, which again depends on the localization.

Sorry I didn't mention that in the first round, I'll quickly fix it myself since it is a bit involved with the localization stuff.

PR #194, opened 2021-04-26
minor checks for list values to prevent errors

jskills:
This may be specific to those who have imported Google Play Music libraries into Youtube Music.  As meticulous as I have been about ensuring all songs have the minimum mp3 tags (song title , artist and genre) as well as "album" wherever possible, I have seen results come back from ytmusicapi searches on "uploads" where the artist or album is missing and cause parsers/browsing.py or parsers/utils.py throw errors.  These searches on the same terms work and return results in Youtube Music in the browser.

These minor additions ensure we're checking for values in Python list structures in some cases where they're assumed to be there.  These changes just make for more graceful performance when encountering data that is expected to be present, but may not be.

### DIFF
--- a/ytmusicapi/parsers/browsing.py
+++ b/ytmusicapi/parsers/browsing.py
@@ -81,8 +81,8 @@ class Parser:
                         search_result['album'] = {
                             'name': flex_items[1][2]['text'],
                             'id': nav(flex_items[1][2], NAVIGATION_BROWSE_ID)
-                        }
-                        search_result['duration'] = flex_items[1][4]['text']
+                        } if flex_items[1].count(2) else ''
+                        search_result['duration'] = flex_items[1][4]['text'] if flex_items[1].count(4) else ''
                     search_result['resultType'] = 'song'
 
                 else:  # artist or album result
@@ -95,7 +95,7 @@ class Parser:
                             run['text'] for i, run in enumerate(flex_item2['text']['runs'])
                             if i % 2 == 0
                         ]
-                        search_result['artist'] = runs[1]
+                        search_result['artist'] = runs[1] if runs.count(1) else ''
                         if len(runs) > 2:  # date may be missing
                             search_result['releaseDate'] = runs[2]
                         search_result['resultType'] = 'album'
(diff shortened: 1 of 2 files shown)


sigma67:
Thanks for the PR! At first glance these seem to be band-aid fixes to not cause errors anymore. That is of course a priority. 

However, ideally I'd like to get to a point where all information on the web interface is also correctly returned by ytmusicapi. Some comments:

- Can you check the correctness of the returned keys with these changes? I.e. does the artist key contain an artist, and not an album for example? This is something that often happens with number-based indexing when the number of items is different from the expected
- I have to object to the change in utils.py since a) the change renders the if-else pointless and b) this causes fail-silent behavior which may cause unexpected errors in other places (esp. since this function is pretty much used everywhere in the API). What specific error were you trying to fix with this change? I'm sure we can find a better solution.

jskills:
Yes - fixes were intended to be minimal and indeed band-aids in order to avoid errors .

I added the output JSON to the Google doc under each of the first 2 example searches.  So you can now see (1) errors using ytmusicapi (2) same search in YTmusic browser and (3) results from ytmusicapi with changes applied (included links to pastebin so you can view the complete JSON output).

[https://docs.google.com/document/d/1SPs0KqeVSYGBtypUfqFFY4hSiNqc8BSMhSF3_MxV9gg/edit](https://docs.google.com/document/d/1SPs0KqeVSYGBtypUfqFFY4hSiNqc8BSMhSF3_MxV9gg/edit)

In regards to the change to utils.py, yes that was rather hasty on my part.  We can indeed do better here.  It seems that the "none_if_absent" parameter default is what I should have offered as a change here.  I could be mistaken, but I don't see none_if_absent being explicitly set anywhere that calls this method.  So would it be acceptable to do this?


ytmusicapi/parsers/utils.py" line 137 ...

def nav(root, items, none_if_absent=True):

sigma67:
Hey, I ended up using a different approach in 391eb3f so I'll close this PR. Your PR and the document were very helpful though in determining what's causing your issues. I'll give due credit in the next release.

Please test with latest master if this fixes your errors.

jskills:
Pulled down the latest master and all searches on uploads come back without errors.

Thanks so much for your fast responses and for taking the time to work through this.

Issue #196, opened 2021-05-17
Is get_streaming_data still working? 


KoljaWindeler:
Hi, this is one of those "some user have a problem but not all" things.

Original issue:  
https://github.com/KoljaWindeler/ytube_music_player/issues/97
I can't reproduce the issue myself but quite a few user suddenly lost the ability to listen to music.

My code would call get_streaming_data but (according to the logs) would only get an empty response {}.

As of now I use a backup method for those users but wonder why get_streaming_data isn't working anymore for so many users.

This is the interesting part:
2021-05-17 12:46:11 ERROR (MainThread) [custom_components.ytube_music_player.media_player] No adaptiveFormat and no formats found
2021-05-17 12:46:11 ERROR (MainThread) [custom_components.ytube_music_player.media_player] get_streaming_data(SQyu9vfW9Bk)
2021-05-17 12:46:11 ERROR (MainThread) [custom_components.ytube_music_player.media_player] {}

https://github.com/KoljaWindeler/ytube_music_player/blob/dc06c8b1c37c6cc0fc01769f7d9854f11a4a4f26/custom_components/ytube_music_player/media_player.py#L1118

sigma67:
It might be related to the x-goog-visitor-id. Can you have your users try again with the latest version? Perhaps that already fixed the issues.

KoljaWindeler:
hah, interesting .. now I have the same issue with 17.0  ... what can i do to get you the needed infos?

sigma67:
Thanks for being persistent, it seems the flow has changed on the YouTube side. There is now a YouTube Music specific endpoint, it no longer uses the YouTube get_video_info.

The new endpoint is `https://music.youtube.com/youtubei/v1/player`. Besides `streamingData `it also returns some other things like `videoDetails`, `playerConfig`, `storyboards` and an `attestation`, to prevent  bots. 

I propose we replace the existing call with that one, but keep the method name.

KoljaWindeler:
I can totally adopt as the current result is empty anyway ;)

sigma67:
I pushed a simple implementation, feel free to try and provide feedback :)

The method is now `get_song `only which also provides the streaming data.

KoljaWindeler:
hmm get_song is also {} .. nothing got returned .. strange, right?

KoljaWindeler:
[...]

PR #198, opened 2021-05-22
Added get_signatureTimestamp() and get_basejs_url() methods for #196.

impliedchaos:
Didn't really know where to put them, so I just shoved 'em into the browsing mixin...

...and I just now noticed @KoljaWindeler's comment about submitting the PR on #196... 🤦‍♂️


sigma67:
Thank you, a very clean implementation. Only changing the return type to `int` to fit with what get_song is expecting.

Issue #205, opened 2021-06-09
How to get the moods and genres located in https://music.youtube.com/moods_and_genres


YashMakan:
I am looking for a way to get the moods and genres. How can I do that?

impliedchaos:
Something like this works:
```
from ytmusicapi import YTMusic
from ytmusicapi.parsers.utils import (nav, NAVIGATION_BROWSE_ID, SECTION_LIST, SINGLE_COLUMN_TAB, TITLE_TEXT)

ytm = YTMusic('auth.json')

# Get moods and genre categories.
response = ytm._send_request("browse", {"browseId": "FEmusic_moods_and_genres"})
moods = {}
for sect in nav(response, SINGLE_COLUMN_TAB + SECTION_LIST):
    for cat in nav(sect, ["gridRenderer", "items"]):
        title = nav(cat, ["musicNavigationButtonRenderer", "buttonText", "runs", 0, "text"]).strip()
        endpnt = nav(cat, ["musicNavigationButtonRenderer", "clickCommand", "browseEndpoint", "browseId"])
        params = nav(cat, ["musicNavigationButtonRenderer", "clickCommand", "browseEndpoint", "params"])
        moods[title] = {"name": title, "params": params, "endpoint": endpnt}

# Print the list of categories.
for m in moods.keys():
    print(m)

# Get playlists from "Focus"
mood = "Focus"
response = ytm._send_request("browse", {"browseId": moods[mood]["endpoint"], "params": moods[mood]["params"]})
for sect in nav(response, SINGLE_COLUMN_TAB + SECTION_LIST):
    key = []
    if "gridRenderer" in sect:
        key = ["gridRenderer", "items"]
    elif "musicCarouselShelfRenderer" in sect:
        key = ["musicCarouselShelfRenderer", "contents"]
    elif "musicImmersiveCarouselShelfRenderer" in sect:
        key = ["musicImmersiveCarouselShelfRenderer", "contents"]
    if len(key):
        for item in nav(sect, key):
            title = nav(item, ["musicTwoRowItemRenderer"] + TITLE_TEXT).strip()
            brId = nav(item, ["musicTwoRowItemRenderer"] + NAVIGATION_BROWSE_ID)
            print("Playlist Title: " + title + " - Playlist ID: " + brId)
```

Probably better off using `find_object_by_key` instead of `nav` all the time, but it's just an example.

sigma67:
[...]

PR #210, opened 2021-07-01
Methods for browsing "moods and genres" playlists for #205

impliedchaos:
- added `get_mood_categories()` to retrieve the "Moods & Genres" categories and sub-categories from YTMusic
- added `get_mood_playlists()` to retrieve a list of playlists from a "Moods & Genres" category


sigma67:
Very nice implementation. For `get_mood_playlists` the current response is missing the section headers (Spotlight and Featured), but I don't think it matters. Good work!

PR #244, opened 2021-12-13
Add zh_CN language files

BruceZhang1993:
language='zh_CN'

changed files

ytmusicapi/locales/zh_CN/LC_MESSAGES/base.mo
ytmusicapi/locales/zh_CN/LC_MESSAGES/base.pot


sigma67:
Hey, thanks for the PR! Can you tell me which language of the multiple Chinese options in the UI you used?

The reason I'm asking is that the "Explicit" labels aren't localized yet, which will need to be done in the future to correctly recognize Explicit labels in different languages. Same goes if we need to add other labels in the future.

image

BruceZhang1993:
@sigma67  zh_CN is for 中文（简体）  the first option in the screenshot

PR #263, opened 2022-03-04
feature/add ko locales

HHongSeungWoo:
add korean translated file


ytmusicapi/helpers.py
```diff
@@ -96,7 +96,7 @@ def get_datestamp():
 
 
 def to_int(string):
-    number_string = re.split('[\x20\xa0]', string)[0]
+    number_string = re.sub('[^\\d]', '', string)
```
sigma67, line 99: Seems reasonable, but is there a reason why you included this with the Korean language PR?

HHongSeungWoo: because, the word order of korean is different from others

PR #265, opened 2022-03-24
Return video type on get watch playlist

lsmacedo:
This PR makes the necessary changes for videoType to be returned on the `get_watch_playlist` function.
It also updates `browsing.py` so the parameter is returned for both songs and videos, therefore we have now three possible values: `MUSIC_VIDEO_TYPE_ATV`, `MUSIC_VIDEO_TYPE_OMV` and `MUSIC_VIDEO_TYPE_UGC`.


sigma67:
Merged in 56b266d

PR #267, opened 2022-04-08
get_artist: Don't fail on channels with only one category of content

JBBgameich:
The current check is meant to detect channels with only video content,
but it also hits small bands that for example have only released singles
so far. An example for this is "UC1uK8RT3m4nNpXmOSsFsZeg".

Since failing is not really helpful anyway (for example audiotub could still play only the sound of a channel that contains only videos),
I propose to remove the check altogether.


sigma67:
This exception was introduced in #181. It seems reasonable in my opinion to filter out invalid channels with a proper exception.

There should be a better way to filter these though than just checking for the length. I agree that the current solution is far from ideal since it affects small artists.

Issue #32, opened 2020-06-24
Add favourite artist to taste


allanin:
You can choose your favourite artist with this url: https://music.youtube.com/tasteprofile

According to Google it changes to recommendations.

Please add a way to add artists to this list via your api.

sigma67:
This requires two new methods `get_tasteprofile` and `set_tasteprofile`, both using the browse endpoint. 

The get uses the browseId `FEmusic_tastebuilder`, while the set uses `FEmusic_home `and takes selected and impression values from the get method as arguments.

YashMakan:
any improvement in this issue? I want to implement this feature.

sigma67:
Feel free, I don't think anyone else is working on it

sigma67:
Contributed by @adrinu in #289

PR #289, opened 2022-08-21
Added favorite artists to taste

adrinu:
### Issue
#32 

### Description
Added ability add favorite artists to https://music.youtube.com/tasteprofile. Implementation was based on the discussion in above issue

### DIFF
--- a/tests/test.py
+++ b/tests/test.py
@@ -165,10 +165,19 @@ class TestYTMusic(unittest.TestCase):
         self.assertIsNotNone(signatureTimestamp)
 
     def test_set_tasteprofile(self):
-        self.assertRaises(self.yt.set_tasteprofile("Galactic"))
+        artists = [artist for artist in self.yt.get_tasteprofile()]
+
+        self.assertRaises(self.yt.set_tasteprofile(artists[:1]))
+        self.assertRaises(self.yt.set_tasteprofile(artists[:5]))
+
+        self.assertRaises(self.yt_auth.set_tasteprofile(artists[:1]))
+        self.assertRaises(self.yt_auth.set_tasteprofile(artists[:5]))
 
     def test_get_tasteprofile(self):
-        result = self.yt.get_tasteprofiles()
+        result = self.yt.get_tasteprofile()
+        self.assertGreaterEqual(len(result), 0)
+
+        result = self.yt_auth.get_tasteprofile()
         self.assertGreaterEqual(len(result), 0)
 
     ################
--- a/tests/test.py
+++ b/tests/test.py
@@ -164,9 +164,16 @@ class TestYTMusic(unittest.TestCase):
         signatureTimestamp = self.yt.get_signatureTimestamp()
         self.assertIsNotNone(signatureTimestamp)
 
-    ###############
+    def test_set_tasteprofile(self):
+        self.assertRaises(self.yt.set_tasteprofile("Galactic"))
+
+    def test_get_tasteprofile(self):
+        result = self.yt.get_tasteprofiles()
+        self.assertGreaterEqual(len(result), 0)
+
+    ################
     # EXPLORE
-    ###############
+    ################
 
     def test_get_mood_playlists(self):
         categories = self.yt.get_mood_categories()
@@ -206,9 +213,9 @@ class TestYTMusic(unittest.TestCase):
         playlist = self.yt_brand.get_watch_playlist_shuffle(playlistId=config['playlists']['own'])
         self.assertEqual(len(playlist['tracks']), 4)
 
-    ###############
+    ################
     # LIBRARY
-    ###############
+    ################
 
     def test_get_library_playlists(self):
         playlists = self.yt_auth.get_library_playlists(50)
--- a/ytmusicapi/mixins/browsing.py
+++ b/ytmusicapi/mixins/browsing.py
@@ -711,29 +711,48 @@ class BrowsingMixin:
 
         return int(match.group(1))
 
-    def get_tasteprofiles(self) -> Dict:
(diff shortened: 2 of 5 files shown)


ytmusicapi/parsers/browsing.py
```diff
@@ -288,3 +287,17 @@ def parse_related_artist(data):
         'subscribers': subscribers,
         'thumbnails': nav(data, THUMBNAIL_RENDERER),
     }
+
+
+def parse_tasteprofiles(data):
```
sigma67, line 292: Not sure this needs to be in parsers/browsing, since it's not being reused anywhere (and likely won't be in the future). Unless you can think of some case where i might be reused, of course.

ytmusicapi/mixins/browsing.py
```diff
@@ -711,3 +710,41 @@ def get_signatureTimestamp(self, url: str = None) -> int:
             raise Exception("Unable to identify the signatureTimestamp.")
 
         return int(match.group(1))
+
+    def get_tasteprofiles(self) -> Dict:
```
sigma67, line 714: I think this should be singular to be more consistent. `get_tasteprofile`

ytmusicapi/mixins/browsing.py
```diff
@@ -711,3 +710,41 @@ def get_signatureTimestamp(self, url: str = None) -> int:
...
+                "selectionValue": "tastebuilder_selection=/m/05mt_q",
+                "impressionValue": "tastebuilder_impression=/m/05mt_q"
+            }
+        }
+        """
+        response = self._send_request('browse', {'browseId': "FEmusic_tastebuilder"})
+        return parse_tasteprofiles(response)
+
+    def set_tasteprofile(self, *artists: str) -> None:
+        """
```
sigma67, line 732: The docstring is missing the artists param and return value.

ytmusicapi/mixins/browsing.py
```diff
@@ -711,3 +710,41 @@ def get_signatureTimestamp(self, url: str = None) -> int:
...
+            "Drake": {
+                "selectionValue": "tastebuilder_selection=/m/05mt_q",
+                "impressionValue": "tastebuilder_impression=/m/05mt_q"
+            }
+        }
+        """
+        response = self._send_request('browse', {'browseId': "FEmusic_tastebuilder"})
+        return parse_tasteprofiles(response)
+
+    def set_tasteprofile(self, *artists: str) -> None:
```
sigma67, line 731: Conventionally ytmusicapi uses List[str] for multiple str values, see for example here: https://github.com/sigma67/ytmusicapi/blob/d66f20aa13ce2117fce7abcadbb9ae288bd352e7/ytmusicapi/mixins/playlists.py#L239-L241

The function should return the YouTube Music response or error, not `None`.

ytmusicapi/mixins/browsing.py
```diff
@@ -711,3 +710,41 @@ def get_signatureTimestamp(self, url: str = None) -> int:
...
+        }
+
+        for artist in artists:
+            if artist not in taste_profiles:
+                raise Exception("The artist, {}, was not present in taste!".format(artist))
+            formData["selectedValues"].append(taste_profiles[artist]["selectionValue"])
+
+        body = {'browseId': "FEmusic_home", "formData": formData}
+
+        self._send_request('browse', body)
```
sigma67, line 750: The function should return the YouTube Music response or error, not `None`.

tests/test.py
```diff
@@ -164,9 +164,16 @@ def test_get_signatureTimestamp(self):
         signatureTimestamp = self.yt.get_signatureTimestamp()
         self.assertIsNotNone(signatureTimestamp)
 
-    ###############
+    def test_set_tasteprofile(self):
+        self.assertRaises(self.yt.set_tasteprofile("Galactic"))
```
sigma67, line 168: The test should also showcase passing in multiple artists and also test with the authenticated profile `yt_auth`.

sigma67 (changes requested):
Thanks for this contribution! Glad someone finally tackled this. 

Already looks pretty good to me. I've added some minor comments where needed, after that I think we're good to merge.

ytmusicapi/mixins/browsing.py
adrinu: The return response does not contain any useful information that a user can use. Error is already handled inside of `self._send_request`. Though, if something must be returned, then an error would fit best.  A try/except block on the `self._send_request` would suffice.

sigma67: I checked the response myself and you are right, it contains merely the redirect to the home page. I don't think we need a try except around _send_request, as it's only throwing HTTP-type exceptions, which are needed. Consider this point resolved

sigma67:
I think we should also add some info on how the tasteprofile works, i.e. that you can only get a list of suggestions, but not the actual list of selected entries from the server (afaik). 

In addition, the new functions need to be added to the docs as well, see here: https://github.com/sigma67/ytmusicapi/blob/master/docs/source/reference.rst#Browsing

adrinu:
Addressed the comments. For future references, should I be resolving the threads or let you resolve them?

sigma67 (approved):

sigma67:
> Addressed the comments. For future references, should I be resolving the threads or let you resolve them?

It's fine the way you did it. If something's wrong, I can always unresolve the conversation.

sigma67:
I didn't quite like the way how `get_tasteprofile` _had_ to be called twice, since users are encouraged to call it before `set_tasteprofile` anyway to avoid the exception. Therefore I added an optional parameter to pass the result of a prior `get_tasteprofile` call.

Issue #298, opened 2022-09-23
get_history() fails


atlasrule:
Function get_history() started returning empty array, while history is not empty. No authentication problems.

Program raises this:
Item 2: 'overlay'
Item 3: 'overlay'
Item 2: 'overlay'
Item 3: 'overlay'
.
.

sigma67:
I'm not seeing this issue on my end. Has the response format changed?

atlasrule:
Hi @sigma67, first thing came to my mind was response format change, haven't checked the code for weeks prompted that my authentication has expired, probably due to logout. Faced the issue after I renewed my credentials. But I can't say if there's any format change.

sigma67:
merged in 7de7ecf

PR #299, opened 2022-09-23
Add check if absent

atlasrule:
Checking if nested objects absent solved solved my issue, #298 I'm not sure if it's suitable.


sigma67:
merged in 7de7ecf

Issue #303, opened 2022-10-03
Proposal: return playlistIds in get_library_albums


zarkdav:
Hi,

I wanted to merge the content of a brand account into the main and tried to use `get_library_albums` with `rate_playlist`, unfortunately the first method was not returning `playlistId` as part of the album dict and the second method was not recognizing the `browseId`.

I have written two small modifications to `navigation.py` and `parsers/library.py` which seem to solve this. Should I go ahead and open a pull request?

sigma67:
That seems quite reasonable to implement, thanks for the suggestion. Feel free to submit a PR! 

Will need to update the docstrings for the affected functions as well.

PR #304, opened 2022-10-03
Add playlistId key to the album dicts in parse_album

zarkdav:
This is the pull request related to Issue #303.

### DIFF
--- a/ytmusicapi/navigation.py
+++ b/ytmusicapi/navigation.py
@@ -60,7 +60,7 @@ MTRIR = 'musicTwoRowItemRenderer'
 TASTE_PROFILE_ITEMS = ["contents", "tastebuilderRenderer", "contents"]
 TASTE_PROFILE_ARTIST = ["title", "runs"]
 SECTION_LIST_CONTINUATION = ['continuationContents', 'sectionListContinuation']
-MENU_PLAYLIST_ID = MENU_ITEMS + [0 + 'menuNavigationItemRenderer'] + NAVIGATION_WATCH_PLAYLIST_ID
+MENU_PLAYLIST_ID = MENU_ITEMS + [0, 'menuNavigationItemRenderer'] + NAVIGATION_WATCH_PLAYLIST_ID
 
 
 def nav(root, items, none_if_absent=False):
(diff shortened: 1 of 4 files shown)


ytmusicapi/navigation.py
```diff
@@ -60,6 +60,7 @@
 TASTE_PROFILE_ITEMS = ["contents", "tastebuilderRenderer", "contents"]
 TASTE_PROFILE_ARTIST = ["title", "runs"]
 SECTION_LIST_CONTINUATION = ['continuationContents', 'sectionListContinuation']
+MENU_PLAYLIST_ID = MENU_ITEMS + [0 + 'menuNavigationItemRenderer'] + NAVIGATION_WATCH_PLAYLIST_ID
```
sigma67, line 63: This part is not actually working. Please test your changes before submission.

`TypeError: unsupported operand type(s) for +: 'int' and 'str'`

sigma67 (changes requested):

sigma67:
It would also be great if you could modify the docstring of `get_library_albums` to include your change.

https://github.com/sigma67/ytmusicapi/blob/feac9501d9e49a09285fa60d919f1b5f97ed1b69/ytmusicapi/mixins/library.py#L100-L119

ytmusicapi/parsers/library.py
```diff
@@ -45,6 +45,7 @@ def parse_albums(results):
         data = result[MTRIR]
         album = {}
         album['browseId'] = nav(data, TITLE + NAVIGATION_BROWSE_ID)
+        album['playlistId'] = nav(data, MENU_PLAYLIST_ID)
```
sigma67, line 48: `none_if_absent` parameter should be set to true to avoid errors if the menu order is changed or the playlistId is missing for some other reason. This was the case at least once in my tests

ytmusicapi/navigation.py
zarkdav: Yes, apologies, I wanted to prettify my change and should have taken the time to check it.

ytmusicapi/parsers/library.py
zarkdav: Good idea! although I would love to get an extract from the json you got in your response in order to try and adapt to this situation.

Issue #294, opened 2022-09-15
Breaking value in Chrome cookie for auth


scrimshander:
Just had to refresh my auth headers after many many months, and kept getting a missing key bug in the cookie, pasted from Chrome/developer tools.

Turns out this key/value: 

`YT_CL={\"loctok\":\"APLXd64Le9G-vJVqzKjy-OsOVBh6NfU2xKnD6awbKB9lJUXzvHGMuNBmjOpK5q_K xVqxgIRcjI_FC27tj-wmyhvUgW8g4A\"};`

Was breaking the cookie parser (perhaps because of the quotes or the embedded dict). Removing it from the headers text fixed the issue.

sigma67:
This seems strange, your cookie value is not escaped properly. Can you please paste the original `YT_CL` headers from Chrome? Excluding the identifying `Cookie` entry and any private information in there, it's not relevant here.

JohnHKoh:
Same issue here - in the Chrome DevTools, the cookie looks like:

```
YT_CL={"loctok":"stringofchars"};
```

Before, I was receiving the error: "Exception: Your cookie is missing the required value __Secure-3PAPISID". But after removing the "YT_CL" cookie, it started to work again.

PR #309, opened 2022-10-13
Remove quotes from cookies

JohnHKoh:
Due to the SimpleCookie issue described [here](https://github.com/sigma67/ytmusicapi/issues/294), cookies are not properly parse when they include quotes. As the "YT_CL" cookie includes quotes, this breaks the cookie parser and throws the following exception:

```
Exception: Your cookie is missing the required value __Secure-3PAPISID
```

While one solution is to [completely re-write the cookie parser](https://github.com/django/django/commit/93a135d111c2569d88d65a3f4ad9e6d9ad291452) to fix this issue, a much simpler one may be to just remove quotes from cookies altogether. Authentication succeeds after removing the quotes from the cookies.

### DIFF
--- a/ytmusicapi/helpers.py
+++ b/ytmusicapi/helpers.py
@@ -42,7 +42,7 @@ def get_visitor_id(request_func):
 
 def sapisid_from_cookie(raw_cookie):
     cookie = SimpleCookie()
-    cookie.load(raw_cookie)
+    cookie.load(raw_cookie.replace("\"", ""))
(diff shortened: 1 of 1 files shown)


sigma67:
Thanks for picking this up and finding the root cause. Seems to be an issue with the Python standard library: https://github.com/python/cpython/issues/92936

Seems sensible to simply strip the quotes here, as we don't need that value for the sapisid.

PR #394, opened 2023-05-18
Unrecognized case sensitive headers during auth

czifumasa:
Browser authentication doesn't work when I copy request headers from Chrome. 
The reason is request headers are Capitalized in Chrome. 

From [CaseInsensitiveDict](https://requests.readthedocs.io/projects/cn/zh_CN/latest/_modules/requests/structures.html) docs:
`All keys are expected to be strings. The structure remembers the
    case of the last key to be set, and ``iter(instance)``,
    ``keys()``, ``items()``, ``iterkeys()``, and ``iteritems()``
    will contain case-sensitive keys. However, querying and contains
    testing is case insensitive::`

With that in mind I replaced "headers.keys()" method with better check and now check for request headers should be case insensitive.


sigma67:
Merged in d9f9fcc

PR #396, opened 2023-05-28
Add external OAuth case to browser check

MarvinSchenkel:
Hi 👋 ,

I use this project as a dependency for a project where I retrieve an OAuth token myself and set it on the headers to authenticate. Since version 1.0.1 this started failing, because this method is not recognized as either browser_auth or oauth. 

This PR extends the browser check to make sure that the `Authorization` value is an OAuth token in case no `Cookie` is supplied.

Regards,
Marvin

### DIFF
--- a/ytmusicapi/auth/browser.py
+++ b/ytmusicapi/auth/browser.py
@@ -10,7 +10,10 @@ path = os.path.dirname(os.path.realpath(__file__)) + os.sep
 
 def is_browser(headers: CaseInsensitiveDict) -> bool:
     browser_structure = {"authorization", "cookie"}
-    return all(key in headers for key in browser_structure)
+    return all(key in headers for key in browser_structure) or (
+        "authorization" in headers and headers["authorization"].startswith("Bearer ")
+    )
+    
 
 
 def setup_browser(filepath=None, headers_raw=None):
--- a/ytmusicapi/auth/browser.py
+++ b/ytmusicapi/auth/browser.py
@@ -10,10 +10,7 @@ path = os.path.dirname(os.path.realpath(__file__)) + os.sep
 
 def is_browser(headers: CaseInsensitiveDict) -> bool:
     browser_structure = {"authorization", "cookie"}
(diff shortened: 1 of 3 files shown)


sigma67:
Just to make sure I'm understanding this correctly. You're using `browser` auth with an OAuth `Bearer` token? And in that case you're not passing the `cookie` attribute?

That seems like a bit of an edge case, although I wouldn't want to prevent it if it works for you. 

Either way I'd rather modify the outer scope to prevent breaking `is_browser` for anyone else:

https://github.com/sigma67/ytmusicapi/blob/d9f9fcc1477ead219e7c06c593574ba8cc13fe52/ytmusicapi/auth/headers.py#L36-L37

MarvinSchenkel:
> Just to make sure I'm understanding this correctly. You're using `browser` auth with an OAuth `Bearer` token? And in that case you're not passing the `cookie` attribute?
> 
> That seems like a bit of an edge case, although I wouldn't want to prevent it if it works for you.
> 
> Either way I'd rather modify the outer scope to prevent breaking `is_browser` for anyone else. I.e. add another `elif` after this:
> 
> https://github.com/sigma67/ytmusicapi/blob/d9f9fcc1477ead219e7c06c593574ba8cc13fe52/ytmusicapi/auth/headers.py#L36-L37

Yep that's correct. The reason is that I do not want to use the manual CLI approach to obtain an OAuth token, since I want to redirect my users to the login page automatically. 

The reason I picked `is_browser` is because I am actually using a browser. I extended the check and made sure it only passes when a Bearer token is passed when the cookie is omitted. This should not break any existing functionality.

Of course, if you'd rather have a new `elif`, then I will be happy to adjust my PR.

sigma67:
Yes, I'd prefer to create it separate. You could also add another method `is_browser_oauth` in `browser.py`, which you then add to the `elif` in `headers.py`

MarvinSchenkel:
@sigma67  Fixed! :)

PR #403, opened 2023-06-19
Oauth refresh on request

zeitmeister:
This fix should update so that for every request to ytmusic the code checks if the access token needs refreshing and refreshes it if that's the case. 

However, the headers object are updated on every request, regardless if the token needs refreshing or not. This is probably not cpu heavy by any means but should perhaps be looked at in another iteration. or when i have more time...

### DIFF
--- a/ytmusicapi/auth/headers.py
+++ b/ytmusicapi/auth/headers.py
@@ -24,13 +24,13 @@ def prepare_headers(
     session: requests.Session,
     proxies: Optional[Dict] = None,
     auth: Optional[str] = None,
+    oauth: Optional[YTMusicOAuth] = None
 ) -> Dict:
     if auth:
         input_json = load_headers_file(auth)
         input_dict = CaseInsensitiveDict(input_json)
 
         if is_oauth(input_dict):
-            oauth = YTMusicOAuth(session, proxies)
             headers = oauth.load_headers(dict(input_dict), auth)
 
         elif is_browser(input_dict):
--- a/ytmusicapi/auth/headers.py
+++ b/ytmusicapi/auth/headers.py
@@ -23,15 +23,13 @@ def load_headers_file(auth: str) -> Dict:
 def prepare_headers(
     session: requests.Session,
     proxies: Optional[Dict] = None,
-    auth: Optional[str] = None,
-    oauth: Optional[YTMusicOAuth] = None
+    input_dict: Optional[CaseInsensitiveDict] = None,
 ) -> Dict:
-    if auth:
-        input_json = load_headers_file(auth)
-        input_dict = CaseInsensitiveDict(input_json)
+    if input_dict:
 
         if is_oauth(input_dict):
-            headers = oauth.load_headers(dict(input_dict), auth)
+            oauth = YTMusicOAuth(session, proxies)
+            headers = oauth.load_headers(dict(input_dict), input_dict['filepath'])
 
         elif is_browser(input_dict):
             headers = input_dict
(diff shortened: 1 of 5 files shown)


sigma67:
You can easily fix this by extending __init__ so that we know if this is an `is_oauth`  instance (set a new instance variable like `is_browser_auth`). Then simply check for the value of that variable. Also you could simply call `load_headers` directly since we know then that it's an `oauth` instance.

ytmusicapi/auth/headers.py
```diff
@@ -24,13 +24,13 @@ def prepare_headers(
     session: requests.Session,
     proxies: Optional[Dict] = None,
     auth: Optional[str] = None,
+    oauth: Optional[YTMusicOAuth] = None
```
sigma67, line 27: instead of adding another parameter and making this function even more confusing, we should change the `auth` parameter to represent the `input_dict` instead of the filepath to `headers_auth`. That should greatly simplify the code. And this would also prevent reading the file twice (`load_headers_file` is called twice in this iteration)

ytmusicapi/ytmusic.py
```diff
@@ -81,7 +83,13 @@ def __init__(self,
         self.proxies = proxies
         self.cookies = {'CONSENT': 'YES+1'}
 
-        self.headers = prepare_headers(self._session, proxies, auth)
+        input_json = load_headers_file(self.auth)
+        self.input_dict = CaseInsensitiveDict(input_json)
+        self.is_oauth_auth = is_oauth(self.input_dict)
+        if self.is_oauth_auth:
+            self.ytmusic_oauth = YTMusicOAuth(self._session, self.proxies)
```
sigma67, line 90: This `if` should not be needed anymore if we change `prepare_headers` to take the `input_dict` as an argument.

ytmusicapi/ytmusic.py
```diff
@@ -121,7 +129,12 @@ def __init__(self,
             except KeyError:
                 raise Exception("Your cookie is missing the required value __Secure-3PAPISID")
 
+
+
     def _send_request(self, endpoint: str, body: Dict, additionalParams: str = "") -> Dict:
+
+        if self.is_oauth_auth and self.ytmusic_oauth is not None:
+            self.headers = self.ytmusic_oauth.load_headers(dict(self.input_dict), self.auth)
```
sigma67, line 137: with my suggestion you can instead call `prepare_headers` here and remove the second condition in the `if`

sigma67 (changes requested):
I think this is pretty close to a good solution, although still some optimizations to be had

ytmusicapi/ytmusic.py
```diff
@@ -81,7 +83,12 @@ def __init__(self,
         self.proxies = proxies
         self.cookies = {'CONSENT': 'YES+1'}
 
-        self.headers = prepare_headers(self._session, proxies, auth)
+        input_json = load_headers_file(self.auth)
```
sigma67, line 86: Please also consider the case where self.auth is None (non-authenticated instances). Tests won't run on this changeset for this reason

zeitmeister:
I hope that the last fix makes the test run. I also don't know why these commits with merges keeps coming. they seem to be from another fork or something. git wouldn't let me push if i didn't pull stuff that i didn't even change in the first place..

anyway, thanks again for your comments. I think they really improved the funcionality.

sigma67:
This does it. Thanks for the persistence. No worry about the commits, they get squashed anyway

PR #413, opened 2023-07-19
Small edit only - removed redundant loop counter

nick42d:
The count in parse_playlist_items is not referenced elsewhere in the function.


sigma67:
Good catch! This project needs more contributors like you :)

PR #421, opened 2023-08-04
Fix for KeyError when searching community_playlists or featured_playlists. Closes #420

### LINKED ISSUE #420: KeyError exception when searching community_playlists or featured_playlists with scope = library
G'day,

Noticed that the function parsers.search._get_param_2 does not handle "community_playlists" or "featured_playlists".
However, parsers.search.get_search_params calls this function with these parameters when scope is set to "library" and therefore we get a KeyError.

Happy to raise a PR if you are able to advise if this should be handled with a documented exception or alternative params sent to innertube.

Cheers Nick

**To Reproduce**
call YTMusic.search('xxx', 'community_playlists', 'library').
result "KeyError: 'community_playlists'"

nick42d:
Fix for KeyError when searching community_playlists or featured_playlists.
Also applies a small lint to tests.py.

### DIFF
--- a/ytmusicapi/mixins/search.py
+++ b/ytmusicapi/mixins/search.py
@@ -147,9 +147,9 @@ class SearchMixin:
 
         if scope == scopes[0] and filter in filters[3:5]:
             raise Exception(
-                "Community Playlists and Featured Playlists filters cannot be set when searching library. "
-                "Please use one of the following filters or leave out the parameter: "
-                "albums, artists, playlists, songs, videos")
+                f"{filter} cannot be set when searching library. "
+                f"Please use one of the following filters or leave out the parameter: "
+                + ', '.join(filters[0:3]+filters[5:]))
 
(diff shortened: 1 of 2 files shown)


ytmusicapi/mixins/search.py
```diff
@@ -143,6 +145,12 @@ def search(self,
                 "No filter can be set when searching uploads. Please unset the filter parameter when scope is set to "
                 "uploads. ")
 
+        if scope == scopes[0] and filter in filters[3:5]:
+            raise Exception(
+                "Community Playlists and Featured Playlists filters cannot be set when searching library. "
+                "Please use one of the following filters or leave out the parameter: "
+                "albums, artists, playlists, songs, videos")
```
sigma67, line 152:
```suggestion
                f"{filter} cannot be set when searching library. "
                f"Please use one of the following filters or leave out the parameter: "
                f"albums, artists, playlists, songs, videos")
```

PR #438, opened 2023-09-25
fixed KeyError in get_charts

ankydv:
I handled 'KeyError', and added new Navigation PATH in that case.

### DIFF
--- a/ytmusicapi/navigation.py
+++ b/ytmusicapi/navigation.py
@@ -26,6 +26,8 @@ PAGE_TYPE = [
 ]
 WATCH_VIDEO_ID = ['watchEndpoint', 'videoId']
 NAVIGATION_VIDEO_ID = ['navigationEndpoint'] + WATCH_VIDEO_ID
+QUEUE_VIDEO_ID = ['queueAddEndpoint','queueTarget','videoId']
+NAVIGATION_VIDEO_ID_2 = MENU_ITEMS+[0]+MENU_SERVICE+QUEUE_VIDEO_ID
 NAVIGATION_PLAYLIST_ID = ['navigationEndpoint', 'watchEndpoint', 'playlistId']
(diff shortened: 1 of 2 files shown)


ytmusicapi/navigation.py
```diff
@@ -26,6 +26,7 @@
 ]
 WATCH_VIDEO_ID = ['watchEndpoint', 'videoId']
 NAVIGATION_VIDEO_ID = ['navigationEndpoint'] + WATCH_VIDEO_ID
+NAVIGATION_VIDEO_ID_2 = ['menu','menuRenderer','items',0,'menuServiceItemRenderer','serviceEndpoint','queueAddEndpoint','queueTarget','videoId']
```
sigma67, line 29: Please use the existing navigation paths MENU_ITEMS, MENU_SERVICE

ytmusicapi/parsers/browsing.py
```diff
@@ -103,9 +103,13 @@ def parse_song_flat(data):
 def parse_video(result):
     runs = nav(result, SUBTITLE_RUNS)
     artists_len = get_dot_separator_index(runs)
+    try:
+        videoId = nav(result, NAVIGATION_VIDEO_ID)
+    except KeyError:
+        videoId = nav(result, NAVIGATION_VIDEO_ID_2)
```
sigma67, line 109:
```suggestion
    videoId = nav(result, NAVIGATION_VIDEO_ID, True)
    if not videoId:
        videoId = nav(result, NAVIGATION_VIDEO_ID_2)
```

sigma67 (changes requested):

ytmusicapi/navigation.py
ankydv: Done. Please review.

Issue #483, opened 2023-12-07
Search has weird albums


polvallverdu:
**Describe the bug**
I've been running an automated application with this library (my async version), and the only flagging that I get is a report that some albums are considered songs (ytmusic doesn't even recognize these as albums, as I mentioned in a past issue), but from the app I can access the album, so there should be a way to get the album id. This occurs when searching without any filter.

Here's what I'm talking about:
```json
{"category": "Albums", "resultType": "song", "title": "Supernova", "album": null, "inLibrary": false, "feedbackTokens": {"add": null, "remove": null}, "videoId": "1d_R6LqadWQ", "videoType": "MUSIC_VIDEO_TYPE_ATV", "duration": null, "year": "2023", "artists": [{"name": "Saiko", "id": "UCN7aCXjfOVY-YOj_OWD-hSQ"}], "isExplicit": false, "thumbnails": [{"url": "https://lh3.googleusercontent.com/R2d9H4Uyx91YfDTLAC-Vl7mGwXc0ZH5QQ4pV-OE9SbqpznDTT1IgT1dvkW9MhpBMPcBi_yC9mxYwfD4=w60-h60-l90-rj", "width": 60, "height": 60}, {"url": "https://lh3.googleusercontent.com/R2d9H4Uyx91YfDTLAC-Vl7mGwXc0ZH5QQ4pV-OE9SbqpznDTT1IgT1dvkW9MhpBMPcBi_yC9mxYwfD4=w120-h120-l90-rj", "width": 120, "height": 120}, {"url": "https://lh3.googleusercontent.com/R2d9H4Uyx91YfDTLAC-Vl7mGwXc0ZH5QQ4pV-OE9SbqpznDTT1IgT1dvkW9MhpBMPcBi_yC9mxYwfD4=w226-h226-l90-rj", "width": 226, "height": 226}, {"url": "https://lh3.googleusercontent.com/R2d9H4Uyx91YfDTLAC-Vl7mGwXc0ZH5QQ4pV-OE9SbqpznDTT1IgT1dvkW9MhpBMPcBi_yC9mxYwfD4=w544-h544-l90-rj", "width": 544, "height": 544}]}
```

When searching for "supernova saiko", this album appears without album id, but a track also appears containing the actual albumid. That's why I call it weird album.

**To Reproduce**
1. Search for "supernova saiko"
2. Check the returned dictionary.

theyak:
Agreed. I think the issue lies in the search parser. 

```python
[...]

PR #486, opened 2023-12-15
Fixed single albums sometimes not being recognized as albums

polvallverdu:
Related to #483

@theyak was right. The issue was related to not recognizing albums being albums

### DIFF
--- a/ytmusicapi/parsers/search.py
+++ b/ytmusicapi/parsers/search.py
@@ -50,8 +50,11 @@ def parse_search_result(data, search_result_types, result_type, category):
     default_offset = (not result_type or result_type == "album") * 2
     search_result = {'category': category}
     video_type = nav(data, PLAY_BUTTON + ['playNavigationEndpoint'] + NAVIGATION_VIDEO_TYPE, True)
-    if not result_type and video_type:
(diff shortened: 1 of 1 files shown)


ytmusicapi/parsers/search.py
```diff
@@ -50,8 +50,11 @@ def parse_search_result(data, search_result_types, result_type, category):
     default_offset = (not result_type or result_type == "album") * 2
     search_result = {'category': category}
     video_type = nav(data, PLAY_BUTTON + ['playNavigationEndpoint'] + NAVIGATION_VIDEO_TYPE, True)
-    if not result_type and video_type:
-        result_type = 'song' if video_type == 'MUSIC_VIDEO_TYPE_ATV' else 'video'
+    if not result_type:
+        if search_result["category"] == "Albums":
```
sigma67, line 54: Hi @polvallverdu , thanks for your contribution. However, this will not work without localization. Note that "category" is not used anymore else for code flow for this reason. Only result_type is localization independent (after delocalization in l.59)

polvallverdu: Ahh, I'll try to investigate it later today.

Issue #487, opened 2023-12-17
Support for alternate OAuth clients


jcbirdwell:
**Is your feature request related to a problem? Please describe.**
There didn't seem to be a way to clearly and easily change the underlying OAuth client being used to retrieve user credentials. This acted as a barrier to any added level of control, scalability, and end-user clarity that comes with using a custom OAuth client. 

**Describe the solution you'd like**
An implementation of YTMusicOAuth that allows for a client id and secret to be provided as parameters, which will be used instead of those found in constants for user login and header creation, and subsequent integration into the YTMusic class.

**Describe alternatives you've considered**
1a. Simply changing the constants: While this solution is certainly the easiest path to the end goal, it's implementation lacks clarity, stability, and flexibility all of which would make it more of a pain than just writing in the feature. 
1b. Ability to change the client id and secret via environmental variables. This was what I went with for the first iteration of the proposed changes, but it ended up not being all that much better in terms of clarity and flexibility in practice.
2. Subclassing of YTMusicOAuth and the ability to provide YTMusic with the OAuth instance directly: While this is probably the top of the line solution, its also the heaviest. The proposed version accomplishes about the same end result, while maintaining the same general structure and flow as the current code and the efficiency boost and simplicity of session sharing at a smaller scale. It also avoids any code duplication which could result from maintaining compatibility with browser based authentication. 
3. Just not doing it: It's a niche issue that adds potentially unwarranted complexity, but I felt the added level of control and utility significantly outweighed this small increase to complexity. Additionally, the changes posture the package in such a way as to be more easily forked/upgraded into a wrapper for the official API, whenever that arrives.

**Additional context**
It was a feature that I wanted for my use-case and therefore something I was going to write anyway. I figured I'd offer it as a contribution instead of just hoarding it.

sigma67:
Hi, I'm in favor of your contribution, although I think we can simplify some things. Continued in #488.

Issue #470, opened 2023-11-09
`get_album_browse_id()` not working for some URLs


bharat-nair:
**Describe the bug**
While using `get_album_browse_id()` , some playlist URLs do not give any `browse_id`

**To Reproduce**
`YTMusic().get_album_browse_id("OLAK5uy_nbMYyrfeg5ZgknoOsOGBL268hGxtcbnDM")` will give an empty string.

**Additional context**
Upon checking the page source for this particular playlist, (https://music.youtube.com/playlist?list=OLAK5uy_nbMYyrfeg5ZgknoOsOGBL268hGxtcbnDM) the `browseId` seems to be there but in this string unescaped form:
`...x22:\x7b\x22browseId\x22:\x22MPREb_scJdtUCpPE2\x22,\x22b...`

bharat-nair:
Going through the implementation, ([L344](https://github.com/sigma67/ytmusicapi/blob/2d2ab999a283200c89cf5a13a3ccf854b3dcc7f2/ytmusicapi/mixins/browsing.py#L344)) perhaps another regex can be added like: `re.findall(r"\\x22MPRE.+?\\x22", response.text)`

sigma67:
Some unicode shenanigans going on, will fix in a moment

PR #488, opened 2023-12-17
Authentication refactor

jcbirdwell:
This is the implementation of #487. It also fixes some issues I ran across while writing the tests for my proposed feature that were preventing a full pass. I can open issues for each or all together if needed, just let me know. Otherwise, they were:
1. A test was throwing a depreciation warning stemming from a regex search in get_album_browse_id, I changed it to use a str.find slicer, which fixed the warning and made it a touch faster.
2. It looks like the songs section of charts is no longer provided by YouTube (could be regional), which was producing an indexing error in get_charts when authenticated. I changed it to check that the results array had enough members before it inserted the songs key. This prevents the error without depreciating the feature should the section be added back on googles end.
3. While prepping a YouTube music account library for the tests I couldn't find the method for saving albums, so I wrote one, only to find afterwards it was written in as a footnote of the rate_playlists method. I went ahead and committed it as an alias/convenience method for other idiots like myself that couldn't find the other. Ignore it if you want. 
4. I may have just been cluelessly butchering my test.py and test.cfg trying to get unitests up and running, but there were a few variables that seemed to be library or region dependent, which I externalized to the cfg so as to be more clearly customizable. If that was a bad call and the purpose of the tests went over my head, let me know so I can redo and rerun with the other variables.


ytmusicapi/mixins/browsing.py
```diff
@@ -341,11 +341,11 @@ def get_album_browse_id(self, audioPlaylistId: str) -> str:
         """
         params = {"list": audioPlaylistId}
         response = self._send_get_request(YTM_DOMAIN + "/playlist", params)
-        matches = re.search(r"\"MPRE.+?\"", response.text.encode("utf8").decode("unicode_escape"))
-        browse_id = None
-        if matches:
-            browse_id = matches.group().strip('"')
-        return browse_id
+
+        if (start := response.text.find('MPREb_')) == -1:
```
sigma67, line 345: I feel like this will break some edge cases and is too simple.

Can you elaborate on the warning you were getting, perhaps we can solve it a different way

ytmusicapi/mixins/explore.py
```diff
@@ -209,9 +209,13 @@ def get_charts(self, country: str = 'ZZ') -> Dict:
             ]))
         charts_categories = ['videos', 'artists']
 
-        has_songs = bool(self.auth)
         has_genres = country == 'US'
         has_trending = country != 'ZZ'
+        # songs section appears to no longer exist, extra length check avoids
```
sigma67, line 214: I suppose the previous check was too simple since you'd actually need to check for a premium subscriber instead of a just a logged-in user. Should probably add a separate account to the tests that is authenticated but not subscribed

So I think we can just remove the `bool(self.auth)` part altogether since it does not actually make a difference

ytmusicapi/mixins/library.py
```diff
@@ -267,6 +267,25 @@ def rate_song(self, videoId: str, rating: str = 'INDIFFERENT') -> Dict:
 
         return self._send_request(endpoint, body)
 
+    def set_album_save(self, audioPlaylistId: str, saved=True):
```
sigma67, line 270: This is identical to [`rate_playlist`](https://github.com/sigma67/ytmusicapi/blob/43d3c3bc0455b73d44235d9d7adf7ad13aea35cc/ytmusicapi/mixins/library.py#L283)

I think it would suffice to add a comment to that method to state that it works on albums as well (if even that).

It is documented clearly in the FAQ: https://ytmusicapi.readthedocs.io/en/stable/faq.html#how-do-i-add-a-song-album-artist-or-playlist-to-my-library

So I would suggest to remove this.

ytmusicapi/ytmusic.py
```diff
@@ -87,12 +92,23 @@ def __init__(self,
...
         if self.auth is not None:
-            input_json = load_headers_file(self.auth)
-            self.input_dict = CaseInsensitiveDict(input_json)
-            self.input_dict['filepath'] = self.auth
+            if isinstance(self.auth, str):
+                input_json = load_headers_file(self.auth)
+                self.input_dict = CaseInsensitiveDict(input_json)
+                self.input_dict['filepath'] = self.auth
+
+            elif isinstance(self.auth, dict):
```
sigma67, line 100: This library is still 3.8 compatible so please fix all instances of `dict`

ytmusicapi/ytmusic.py
```diff
@@ -87,12 +92,23 @@ def __init__(self,
...
+                self.input_dict = self.auth
+
             self.is_oauth_auth = is_oauth(self.input_dict)
 
-        self.headers = prepare_headers(self._session, proxies, self.input_dict)
+        # use custom oauth client parameters if provided
+        if alt_oauth and isinstance(alt_oauth, dict):
+            self.alt_oauth = YTMusicOAuth(self._session, proxies, **alt_oauth)
```
sigma67, line 107: Here you are assuming that the `alt_oauth` dict contains the correct keys for the function, which are documented nowhere except the `YTMusicOAuth.__init__`. Not very transparent.

I suggest we change the `alt_oauth` argument to `YTMusic` to `oauth_credentials` of type `OAuthCredentials`, which can be a new dataclass located in `oauth.py`. It has fields `client_id` and `secret` with default values set to the existing constants. That should simplify things

Instead of calling `YTMusicOAuth` in `ytmusic.py`, just do it in `prepare_headers` by passing in the `OAuthCredentials`.

ytmusicapi/auth/headers.py
```diff
@@ -20,16 +20,26 @@ def load_headers_file(auth: str) -> Dict:
     return input_json
 
 
-def prepare_headers(
-    session: requests.Session,
-    proxies: Optional[Dict] = None,
-    input_dict: Optional[CaseInsensitiveDict] = None,
-) -> Dict:
+def prepare_headers(auth_session: requests.Session | YTMusicOAuth,
```
sigma67, line 23: I think instead of auth_session it would be much nicer and much more readable to just pass in the extra credential for `YTMusicOAuth`

ytmusicapi/ytmusic.py
```diff
@@ -28,16 +28,17 @@ class YTMusic(BrowsingMixin, SearchMixin, WatchMixin, ExploreMixin, LibraryMixin
     """
 
     def __init__(self,
-                 auth: str = None,
+                 auth: Optional[str | dict] = None,
```
sigma67, line 31: I assume this change is only to avoid having to pass a JSON string instead of the dictionary. Not strictly needed,  just nice to have?

sigma67 (changes requested):
Hi @jcbirdwell , thanks so much for contributing this back. Glad to accept this contribution with a few changes.

Please also add documentation in the Setup section (new chapter in Oauth section). How could one obtain such custom credentials?

Requesting a review from @MarvinSchenkel who contributed #396

I will wait for these changes before running the tests myself, I assume you have done so on your machine

heisen273:
Tested this PR and I think all requests through mixin's are broken._. 

at first, `__init__()` goes OK and i can obtain refresh token with my custom client id & secret pair: 
 
Screenshot 2023-12-17 at 23 39 12


But then, when i do any request(`.get_me()`, `.get_playlist()`, etc) - it always returns 401: 

Screenshot 2023-12-17 at 23 39 31


---

I think you have to update `_send_request()` with logics that handles `if self.alt_oauth: ...`

ytmusicapi/auth/oauth.py
```diff
@@ -30,13 +30,35 @@ def is_custom_oauth(headers: CaseInsensitiveDict) -> bool:
...
+            underlying oauth client.
+        """
+
+        if not isinstance(oauth_client_id, type(oauth_client_secret)):
+            raise KeyError(
+                'Alternate oauth credential usage requires an id AND a secret. Pass both or neither.'
+            )
+
+        self.oauth_id = oauth_client_id if oauth_client_id else OAUTH_CLIENT_ID
+        self.oauth_secret = oauth_client_secret if oauth_client_secret else OAUTH_CLIENT_SECRET
```
heisen273, line 54: Was thinking about this overnight, and i think it's overcomplicated approach – instead of introducing new parameters(that has to be always kept in mind & carried over everywhere), why not make `OAUTH_CLIENT_ID` & `OAUTH_CLIENT_SECRET` to load themselves from env variables with fallback to default one's that @sigma67 dedicated for this project & hardcoded in `constants.py`?

ytmusicapi/mixins/browsing.py
jcbirdwell: When running "coverage run -m unittest test.py" from the tests directory. 

```
(python312) birdwell@JCBMPro tests % coverage run -m unittest test.py
s..../Users/birdwell/PycharmProjects/ytmusicapi-contrib/ytmusicapi/ytmusicapi/mixins/browsing.py:345: DeprecationWarning: invalid escape sequence '\/'
  matches = re.search(r"\"MPRE.+?\"", response.text.encode("utf8").decode("unicode_escape"))
....
```

I can't seem to reproduce it outside of coverage/unittest

ytmusicapi/mixins/explore.py
jcbirdwell: I was indeed performing all tests using non-premium accounts.

ytmusicapi/mixins/browsing.py
sigma67: Yes, I see it here on CI as well. It's pretty strange, because the same regex works in a simple isolated script without issue

ytmusicapi/ytmusic.py
jcbirdwell: I'm in favor of a dataclass implementation of OAuthCredentials.

ytmusicapi/ytmusic.py
jcbirdwell: Correct. It enables you to skip dumping a dict to json string only to have it immediately loaded back to a dict.

sigma67: ok, just wanted to make sure I understand the intent

ytmusicapi/auth/oauth.py
jcbirdwell: While I definitely agree on its complexity, I mentioned pursuing environmental variables briefly in the original issue #487 under alternative 1b. IMO, if the feature is something worth implementing, its worth implementing in a way that fully realizes its push for flexibility. My core disagreement with the env approach is that it hampers your ability to switch between clients within a process and also lacks clarity as to which client is currently active. 

That being said, I still like and agree with the utility that an option for using environmental variables provides. What are your thoughts on implementing them as an override within the dataclass mentioned above? IE the OAuthCredentials dataclass default values would be the environmental variables falling back to the dedicated constants when unavailable.

sigma67: Let's not start using environment variables :)

jcbirdwell:
I ended up going with a mix of my original issue solution alternative 2 and the dataclass implementation mentioned previously. While I was adding a credential dataclass I was reminded of why I was originally trying to pass the YTMusicOAuth instance from YTMusic, it was to prevent recreating the OAuth class on every single request. I believe the original intent was to ensure that the authorization field of the header was always up-to-date, this version accomplishes that but instead of rebuilding all the headers each time it just does authorization/ones that update. It also catches #492 in the crossfire and I believe the issue @heisen273 was having with mixins, but I couldn't replicate it so I'm unsure. It's a decent chunk of changes, but should act as a drop-in replacement. I added custom oauth handling and believe it maintains compatibility with #396, but I don't think there is a test for it. I haven't written any of the docs or additional feature tests yet as I wanted to get feedback before tackling em, but I'm getting a full pass for existing tests. In light of that, I'd convert the pull request to a draft, but I'm not positive what that does and don't want to mess anything up.

ytmusicapi/setup.py
```diff
@@ -25,7 +25,10 @@ def setup(filepath: str = None, headers_raw: str = None) -> Dict:
 def setup_oauth(filepath: str = None,
                 session: requests.Session = None,
                 proxies: dict = None,
-                open_browser: bool = False) -> Dict:
+                open_browser: bool = False,
+                env_credentials: bool = False,
```
sigma67, line 29: Please remove this, as stated I am opposed to the introduction of environment variables. This is not a concern for this library

sigma67 (commented):
I'm not sure this second changeset was needed based on what you wanted to achieve. But I like the improvements in general, so let's try to make it work.

I have a very hard time verifying that everything is still working, considering that 18 tests are breaking for me locally even after updating the test.cfg (those changes are quite welcome, thanks).

I suspect most failures have the same origin, some sort of (breaking?) change to the local oauth.json file that I have on my machine. I have attached my test log, maybe you can point out what's wrong.
[tests.log](https://github.com/sigma67/ytmusicapi/files/13719268/tests.log)


I will also state here that this MR should be free of any breaking changes to the public facing API.

Don't worry about draft mode, it has no effect in this repo (since we're currently not running tests on PRs due to somewhat sensitive nature of the credentials)

sigma67 (commented):
The error message has changed slightly, but I'm still getting this:

```
======================================================================
ERROR: test_search_library (test.TestYTMusic.test_search_library)
----------------------------------------------------------------------
  File "\ytmusicapi\tests\test.py", line 164, in test_search_library
    results = self.yt_oauth.search(config['queries']['library_any'], scope="library")
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\mixins\search.py", line 165, in search
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\ytmusic.py", line 200, in _send_request
    if 'X-Goog-Visitor-Id' not in self.headers:
                                  ^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\ytmusic.py", line 191, in headers
    self._headers['authorization'] = self._token.as_auth()
                                     ^^^^^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\auth\oauth.py", line 85, in as_auth
    return f'{self.token_type} {self.access_token}'
                                ^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\auth\oauth.py", line 178, in access_token
    self.token = self.credentials.refresh_token(self.token.refresh_token)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\auth\oauth.py", line 141, in refresh_token
    return OAuthToken.from_response(response)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\ytmusicapi\ytmusicapi\auth\oauth.py", line 30, in from_response
    return cls(**data)
           ^^^^^^^^^^^
TypeError: OAuthToken.__init__() missing 1 required positional argument: 'refresh_token'
```

ytmusicapi/auth/oauth.py
```diff
@@ -22,81 +22,186 @@ def is_oauth(headers: CaseInsensitiveDict) -> bool:
...
+    def from_file(cls, file_path: str, credentials: Optional[OAuthCredentials] = None, sync=True):
+        if os.path.isfile(file_path):
+            with open(file_path) as json_file:
+                file_pack = json.load(json_file)
+
+        creds = credentials if credentials else OAuthCredentials()
+
+        return cls(OAuthToken(**file_pack), creds, file_path if sync else None)
+
+    def __init__(self,
```
sigma67, line 156: the subclass ``__init__`` shouldn't differ from the superclass' ``__init__``

ytmusicapi/auth/oauth.py
```diff
@@ -22,81 +22,186 @@ def is_oauth(headers: CaseInsensitiveDict) -> bool:
...
 
-    def refresh_token(self, refresh_token: str) -> Dict:
+    def prompt_for_token(self, open_browser: bool = False, to_file: Optional[str] = None):
+        code = self.get_code()
+        url = f"{code['verification_url']}?user_code={code['user_code']}"
+        if open_browser:
+            webbrowser.open(url)
+        input(f"Go to {url}, finish the login flow and press Enter when done, Ctrl-C to abort")
+        raw_token = self.token_from_code(code["device_code"])
+        ref_token = RefreshingToken(raw_token, credentials=self)
```
sigma67, line 127: Instantiating a subclass from inside the superclass is an antipattern

ytmusicapi/auth/oauth.py
```diff
@@ -22,81 +22,186 @@ def is_oauth(headers: CaseInsensitiveDict) -> bool:
...
-    """Checks whether the headers contain a Bearer token, indicating a custom OAuth implementation."""
-    return "authorization" in headers and headers["authorization"].startswith("Bearer ")
-
-
-class YTMusicOAuth:
-    """OAuth implementation for YouTube Music based on YouTube TV"""
-
-    def __init__(self, session: requests.Session, proxies: Dict = None):
-        self._session = session
+class OAuthToken:
```
sigma67, line 25: since we have multiple classes now, we can easily split this across 3 files in an `oauth` folder to keep it modular

ytmusicapi/mixins/browsing.py
sigma67: your proposed approach does not work with `OLAK5uy_nbMYyrfeg5ZgknoOsOGBL268hGxtcbnDM` from #470 and would thus introduce a regression

sigma67: If you don't mind I would keep the warning for now and focus on the main goal of the PR

tests/test.py
```diff
@@ -69,6 +76,14 @@ def test_setup_oauth(self, session_mock, json_mock):
             string_headers = headers.read()
             self.yt_oauth = YTMusic(string_headers)
 
+    def test_alt_oauth(self):
+        # ensure client works/ignores alt if browser credentials passed as auth
+        self.assertFalse(self.yt_alt_oauth.is_alt_oauth)
+        # oauth token dict entry and alt
+        self.yt_alt_oauth = YTMusic(json.loads(config['auth']['oauth_token']),
+                                    oauth_credentials=alt_oauth_creds)
+        self.assertTrue(self.yt_alt_oauth.is_alt_oauth)
```
sigma67, line 85: This assert is still failing for me.  The key `expires_at` is missing when `is_oauth` is called, therefore it never goes into the if block to set `is_alt_oauth`

ytmusicapi/ytmusic.py
```diff
@@ -68,10 +70,19 @@ def __init__(self,
         :param location: Optional. Can be used to change the location of the user.
             No location will be set by default. This means it is determined by the server.
             Available languages can be checked in the FAQ.
+        :param oauth_credentials: Optional. Used to specify a different oauth client to be
+            used for authentication flow.
         """
+        self._base_headers = None
```
sigma67, line 76: Please group/order the instance variables and add a docstring via `#: this is what it's for`

ytmusicapi/auth/oauth/credentials.py
```diff
@@ -0,0 +1,84 @@
...
+        "access_token",
+        "expires_at",
+        "expires_in",
+        "token_type",
+        "refresh_token",
+    }
+    return all(key in headers for key in oauth_structure)
+
+
+class OAuthCredentials(Credentials):
```
sigma67, line 26: Please add docstrings to this class as this is where the main logic resides

ytmusicapi/auth/oauth/refreshing.py
```diff
@@ -0,0 +1,76 @@
+from typing import Optional, Any
+import time
+import os
+import json
+
+from .base import OAuthToken, Token, Credentials
+
+
+class RefreshingToken(Token):
```
sigma67, line 9: Please add docstrings here also

ytmusicapi/mixins/explore.py
sigma67: So we can remove `bool(self.auth) and` ?

ytmusicapi/auth/oauth/models.py
```diff
@@ -0,0 +1,29 @@
...
+
+class BaseTokenDict(TypedDict):
+    """ Limited token. Does not provide a refresh token. Commonly obtained via a token refresh. """
+    access_token: str
+    expires_in: int
+    scope: Union[str, DefaultScope]
+    token_type: Union[str, Bearer]
+
+
+class FullTokenDict(BaseTokenDict):
```
sigma67, line 17: Not really a great name tbh. Better to follow the naming of `RefreshingToken`? So `RefreshingTokenDict`?

ytmusicapi/auth/oauth/models.py
```diff
@@ -0,0 +1,29 @@
...
+    """ Entire token. Including refresh. Obtained through token setup. """
+    expires_at: int
+    refresh_token: str
+
+
+class CodeDict(TypedDict):
```
sigma67, line 23: Maybe `AuthCodeDict` ? `Code` is very generic

ytmusicapi/auth/oauth/refreshing.py
```diff
@@ -0,0 +1,76 @@
+from typing import Optional, Any
```
sigma67, line 1: return type hints are missing in this file

ytmusicapi/auth/oauth/base.py
```diff
@@ -0,0 +1,96 @@
+from typing import Optional, Any, Union, Dict
```
sigma67, line 1: return value type hints are missing in this file (except for `Credentials` class)

ytmusicapi/auth/oauth/refreshing.py
```diff
@@ -0,0 +1,76 @@
...
+    def expires_in(self):
+        return None
+
+    @property
+    def scope(self):
+        return self.token.scope
+
+    @property
+    def token_type(self):
+        return self.token.token_type
```
sigma67, line 73: l.51 through l. 73 can be removed as they are not used anywhere, unless I'm missing something

Perhaps a leftover from the refactor

ytmusicapi/ytmusic.py
```diff
@@ -16,7 +17,7 @@
 from ytmusicapi.mixins.library import LibraryMixin
 from ytmusicapi.mixins.playlists import PlaylistsMixin
 from ytmusicapi.mixins.uploads import UploadsMixin
-from ytmusicapi.auth.oauth import YTMusicOAuth, is_oauth
+from ytmusicapi.auth.oauth import OAuthCredentials, is_oauth, RefreshingToken, OAuthToken
 
 
 class YTMusic(BrowsingMixin, SearchMixin, WatchMixin, ExploreMixin, LibraryMixin, PlaylistsMixin,
```
sigma67, line 23: Not related to this line, but please add/update documentation in `docs/source/setup`

sigma67 (changes requested):
Sorry for all the comments, I'd consider this to be the final review hopefully.

Tests are passing for me. 

This one's come a long way and I think it's in a great state now.

Really need to add a proper linter like ruff and some sort of basic no-oauth PR pipeline...

jcbirdwell:
> Sorry for all the comments, I'd consider this to be the final review hopefully.

 I really appreciate all the continued feedback on this PR and I'm satisfied with the direction we indeed up taking as a result. I'll be busy for the next few days, but it shouldn't take long to wrap up the final stuff when I can get back on it.

ytmusicapi/auth/oauth/refreshing.py
```diff
@@ -20,57 +36,81 @@ def __init__(self,
...
-        self.token = token
-        self.credentials = credentials
+        """
+        :param token: Underlying Token being maintained.
+        :param credentials: OAuth client being used for refreshing.
+        :param local_cache: Optional. Path to json file where token values are stored.
+            When provided, file contents is updated upon token refresh.
+        """
+
+        self.token: OAuthToken = token  # internal token being used / refreshed / maintained
```
sigma67, line 46: turn all comments into instance variable docstrings

ytmusicapi/auth/oauth/models.py
```diff
@@ -2,28 +2,32 @@
```
sigma67, line 2: turn into proper docstrings

ytmusicapi/auth/oauth/credentials.py
```diff
@@ -2,50 +2,76 @@
...
+        :param session: Optional. Connection pooling with an active session.
+        :param proxies: Optional. Modify the session with proxy parameters.
+        """
+        # id, secret should be None, None or str, str
+        if not isinstance(client_id, type(client_secret)):
+            raise KeyError(
+                'OAuthCredential init failure. Provide both client_id and client_secret or neither.'
+            )
+
+        # bind instance to OAuth client for auth flows
```
sigma67, line 38: turn into proper docstrings

sigma67 (changes requested):
Thanks for these changes again. It would be great if you could turn all the comments into proper docstrings (most of the comments I added are just todo items regarding this):

https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#directive-autoproperty

See the class example here for properly documenting instance variables.

You can use either style - just add a colon to the current comments or place them in the line above with `#:` (the latter is my preferred style)

Other than that there is only one finding left regarding a try-catch you added

ytmusicapi/auth/oauth/refreshing.py
sigma67: this has not been resolved

sigma67 (approved):

sigma67:
Hi @jcbirdwell , I fixed the remaining issues & merged. Feel free to add the documentation in a separate PR

Issue #509, opened 2024-01-09
Return actual track positions


jcbirdwell:
**Is your feature request related to a problem? Please describe.**
Occasionally some albums will appear on YouTube music with tracks completely missing (not just unavailable). In these cases using the index of the returned track list is insufficient for positioning all tracks. The correct indexing for these tracks is still maintained on YouTube, but the value isn't parsed from the response. 

**Describe the solution you'd like**
Add the track_position key to songs parsed in parse_playlist_items and its corresponding lookup path to the response value.

**Describe alternatives you've considered**
1. Ignore it (never).
2. Separate method for pulling true indexes (seems silly when the value is already there)
3. The path to a song's index in the response is similar to other keys (duration) and provides potential for refactoring out something similar to navigation.nav for extracting these values. I left it as a basic index lookup for now (similar to nearby stuff) as I'm currently unfamiliar with the labyrinth of parsing code and didn't want to dig into it. 

**Additional context**
The album that originally threw the error in my code while using native indexing (browseId: MPREb_TPH4WqN5pUo) https://music.youtube.com/playlist?list=OLAK5uy_n7Xde1CFVX7Y8PeyGiAc2V4ieTRs3PML8

It seems to be a exceedingly rare occurrence.

PR #510, opened 2024-01-09
Track positions

jcbirdwell:
Implementation of #509. 

I refactored to relative imports so I could get proper module linking to track down the required functions.

All relevant tests passing + new tests added for the weird album.
- I was unsure if the album's browseId should be made a fixture or just added inline for the individual test as there seems to be both.

I'm getting some weird behavior with the new ruff/mypy pipeline related to library stubs that I can't seem to get working at the moment, but I think the errors are irrelevant to the PR.


ytmusicapi/__init__.py
```diff
@@ -1,7 +1,7 @@
 from importlib.metadata import PackageNotFoundError, version
 
-from ytmusicapi.setup import setup, setup_oauth
```
sigma67, line 3: Please undo the import changes and try to keep the changeset to the feature request. In general, it is a good idea to keep changesets minimal to ensure smooth sailing and quick merges

ytmusicapi/parsers/playlists.py
```diff
@@ -79,6 +79,12 @@ def parse_playlist_items(results, menu_entries: Optional[List[List]] = None):
             True,
         )
 
+        track_position = None
+        if "index" in data:
+            # similar item nesting to duration above -> both could be refactored into
```
sigma67, line 84: As you noted, please use `nav` here to access the index. You can set `none_if_absent` to simply get `None` if the key is missing (which also catches the case of any subkeys under "index" missing).

ytmusicapi/parsers/playlists.py
```diff
@@ -91,7 +97,9 @@ def parse_playlist_items(results, menu_entries: Optional[List[List]] = None):
             "isExplicit": isExplicit,
             "videoType": videoType,
             "views": views,
+            "track_position": track_position,
```
sigma67, line 100: Not sure about the best key to use here. `track_position`, `track_index`, `track_number`?

Spotify uses `track_number`: https://developer.spotify.com/documentation/web-api/reference/get-an-album

Not sure if we have any other similar examples in `ytmusicapi` - should avoid inconsistencies. AFAIK the index is currently not returned on `get_playlist` either

What do you think?

sigma67 (changes requested):
Good change 👍 Just a few comments.

Also make sure to run pre-commit to fix the failing checks

ytmusicapi/__init__.py
sigma67: Not sure what you meant by "get proper module linking to track down the required functions".

If you configure `ytmusicapi` as the module/source folder in your IDE the function navigation should work without issue, at least it does for me in Pycharm

jcbirdwell: sure thing.

ytmusicapi/parsers/playlists.py
jcbirdwell: I don't feel strongly pulled in any direction. I was fairly indecisive when originally picking the key before settling on track_position. Personally, I've never liked track_number from Spotify's API, but there's definitely an argument to be made for it being less ambiguous here. 

In regards to playlist return, I figured it would just return something similar for playlists, but looking at it now, I don't think the response includes indexing for actual playlists (at least from what's visible on the UI). It might be worth adding an is_album param to parse_playlist_items to avoid a nested track_position lookup that'll never return anything. Thoughts?

sigma67: apple also uses `trackNumber`: https://developer.apple.com/documentation/applemusicapi/songs/attributes

let's not be the odd ones out here

sigma67: and you're right, sadly playlists don't include `trackNumber` - so it makes sense to do as you did

Issue #515, opened 2024-01-13
track_number fails on disabled tracks


jcbirdwell:
**Describe the bug**
Calling get_album on an album containing a disabled, but not removed or deleted, track throws a value error. This is the result of the response having all the keys for the index (nav won't return None) but with a blank space value ' ' instead of the actual track number. 

**To Reproduce**
get_album('MPREb_YuigcYm2erf')

**Additional context**
Unsure if its a regionally disabled track, but it's reproducible with an unauthorized client with location set to 'US'.
Album playlist link for visual context: https://music.youtube.com/playlist?list=OLAK5uy_nQ50qK0OQks_zc_jjn0IvWb7Vg1vo0s5Y

PR #516, opened 2024-01-13
get_album: track_number returns none when track unavailable + test

jcbirdwell:
Simple fix for #515 and adds test case for albums with unavailable tracks. 

Issue could alternatively be solved by extending nav's none_if_absent to coerce nullish return values. thoughts?

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,18 +24,6 @@ def fixture_sample_album() -> str:
     return "MPREb_4pL8gzRtw1p"
 
 
-@pytest.fixture(name="missing_indices_album")
-def fixture_missing_indices_album() -> str:
-    """Sean Paul - Dutty Classics Collection"""
-    return "MPREb_TPH4WqN5pUo"
-
-
-@pytest.fixture(name="disabled_indices_album")
-def fixture_disabled_indices_album() -> str:
-    """Kristian Conde - Absolutely Kristian Conde"""
-    return "MPREb_YuigcYm2erf"
-
-
 @pytest.fixture(name="sample_video")
 def fixture_sample_video() -> str:
     """Oasis - Wonderwall"""
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -24,10 +24,16 @@ def fixture_sample_album() -> str:
     return "MPREb_4pL8gzRtw1p"
 
 
-@pytest.fixture(name="badly_indexed_album")
-def fixture_badly_indexed_album() -> str:
+@pytest.fixture(name="missing_indices_album")
+def fixture_missing_indices_album() -> str:
     """Sean Paul - Dutty Classics Collection"""
     return "MPREb_TPH4WqN5pUo"
+
+
+@pytest.fixture(name="disabled_indices_album")
+def fixture_disabled_indices_album() -> str:
+    """Kristian Conde - Absolutely Kristian Conde"""
+    return "MPREb_YuigcYm2erf"
 
 
 @pytest.fixture(name="sample_video")
--- a/ytmusicapi/parsers/playlists.py
+++ b/ytmusicapi/parsers/playlists.py
@@ -94,8 +94,11 @@ def parse_playlist_items(results, menu_entries: Optional[List[List]] = None, is_
         }
 
         if is_album:
-            track_idx_found = nav(data, ["index", "runs", 0, "text"], True)
-            song["track_number"] = track_idx_found if track_idx_found is None else int(track_idx_found)
+            if not isAvailable:
+                song["track_number"] = None
+            else:
+                # returning None from nav is pointless as the keys are always present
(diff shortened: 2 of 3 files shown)


tests/conftest.py
```diff
@@ -24,12 +24,18 @@ def fixture_sample_album() -> str:
...
-@pytest.fixture(name="badly_indexed_album")
-def fixture_badly_indexed_album() -> str:
+@pytest.fixture(name="missing_indices_album")
+def fixture_missing_indices_album() -> str:
     """Sean Paul - Dutty Classics Collection"""
     return "MPREb_TPH4WqN5pUo"
 
 
+@pytest.fixture(name="disabled_indices_album")
+def fixture_disabled_indices_album() -> str:
```
sigma67, line 34: I applaud the effort but I don't think we need to make every browseId under test a fixture. Just use the id where you need it. If you want to keep it to improve readability, I suggest moving it to the file where it's used -- test_browsing in this case

ytmusicapi/parsers/playlists.py
```diff
@@ -94,8 +94,11 @@ def parse_playlist_items(results, menu_entries: Optional[List[List]] = None, is_
         }
 
         if is_album:
-            track_idx_found = nav(data, ["index", "runs", 0, "text"], True)
-            song["track_number"] = track_idx_found if track_idx_found is None else int(track_idx_found)
+            if not isAvailable:
+                song["track_number"] = None
+            else:
+                # returning None from nav is pointless as the keys are always present
+                song["track_number"] = int(nav(data, ["index", "runs", 0, "text"]))
```
sigma67, line 101:
```suggestion
            song["track_number"] = int(nav(data, ["index", "runs", 0, "text"])) if isAvailable else None
```

a bit more readable imo

sigma67: i wouldn't touch nav because it's got like 200 usages across the project and side effects are not unlikely

tests/conftest.py
jcbirdwell: Sure thing. Was confused on the mixed usage of fixtures and inline ids, but I'll keep it in mind for future changes.

jcbirdwell:
Side note on the new pipeline, these code coverage checks that keep failing. Am I supposed to be doing something differently? My local tests are passing. 

Also I've yet to get mypy to run correctly, it always throws errors about missing type stubs related to requests, despite having installed them via pip and mypy directly. Any recommendations?

```
ytmusicapi/auth/oauth/credentials.py:5: error: Library stubs not installed for "requests"  [import-untyped]
ytmusicapi/auth/oauth/token.py:8: error: Library stubs not installed for "requests.structures"  [import-untyped]
ytmusicapi/mixins/_protocol.py:4: error: Library stubs not installed for "requests"  [import-untyped]
ytmusicapi/auth/browser.py:5: error: Library stubs not installed for "requests.structures"  [import-untyped]
ytmusicapi/mixins/uploads.py:5: error: Library stubs not installed for "requests"  [import-untyped]
ytmusicapi/setup.py:6: error: Library stubs not installed for "requests"  [import-untyped]
ytmusicapi/setup.py:6: note: Hint: "python3 -m pip install types-requests"
ytmusicapi/setup.py:6: note: (or run "mypy --install-types" to install all missing stub packages)
ytmusicapi/setup.py:6: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
ytmusicapi/ytmusic.py:10: error: Library stubs not installed for "requests"  [import-untyped]
ytmusicapi/ytmusic.py:12: error: Library stubs not installed for "requests.structures"  [import-untyped]
```

sigma67:
Don't worry about the test/coverage check, it cannot work yet for external contributors. Right now it only works for people with access to the secrets of this repo, because that's how GitHub does Pull Requests. To get it to work you'd have to set the following variables in *your* fork (just copy paste the text of the equivalent files you use locally): 

image

Normally you'd only have to run ``mypy --install-types`` once, then run it again. Are you running through precommit?

jcbirdwell:
> To get it to work you'd have to set the following variables in _your_ fork (just copy paste the text of the equivalent files you use locally):

Alright cool. I'll give that a whirl.

> Normally you'd only have to run `mypy --install-types` once, then run it again. Are you running through precommit?

Yea I've run `mypy --install-types` and tried installing the stub directly with pip, still throws. Yea I'm running via precommit. 
I ran across something similar as an open issue on the mypy git https://github.com/python/mypy/issues/16400 and was just wondering if you had something setup differently on your machine to avoid it. Outside of trying to install the stubs I setup/configured the environment with pdm/pipx as per the contributing.rst so I'm unsure where the difference would be.

sigma67:
Seems to be a common issue with the precommit mypy action, I remember having some trouble with it as well. Try this: https://stackoverflow.com/a/73603491/5726546

Feel free to update the file in a PR if it works for you

PR #517, opened 2024-01-13
Remove offset + fix top result

AustinGitHub:
Issue - https://github.com/sigma67/ytmusicapi/issues/518
These offsets cause an issue 

Fix, I test this way 
image

Before
[BeforeChange.txt](https://github.com/sigma67/ytmusicapi/files/13929569/BeforeChange.txt)

After
[AfterChanges.txt](https://github.com/sigma67/ytmusicapi/files/13929572/AfterChanges.txt)

Differences
image

Only issue is this 

image

But this is because code forced in to append `None` for artist without `Id` but then again, what if an artist name is "Album" can't just filter that out by keyword. I think this append of `None` should be taken out because with the way youtube music and youtube topic channels are auto-created I think it is safe to say most artists would have an id, no? 

Let me know and I can commit that change, thanks!

### DIFF
--- a/ytmusicapi/parsers/search.py
+++ b/ytmusicapi/parsers/search.py
@@ -35,13 +35,8 @@ def parse_top_result(data, search_result_types):
 
     if result_type in ["song", "video", "album"]:
         on_tap = data.get("onTap")
-        if on_tap is not None:
-            if "watchEndpoint" in on_tap:
-                if nav(on_tap, WATCH_VIDEO_ID) is not None:
-                    search_result["videoId"] = nav(on_tap, WATCH_VIDEO_ID)
-            if "navigationEndpoint" in on_tap:
-                if nav(on_tap, NAVIGATION_VIDEO_TYPE) is not None:
-                    search_result["videoType"] = nav(on_tap, NAVIGATION_VIDEO_TYPE)
+        search_result["videoId"] = nav(data, ["onTap"] + WATCH_VIDEO_ID, True)
+        search_result["videoType"] = nav(data, ["onTap"] + NAVIGATION_VIDEO_TYPE, True)
 
         search_result["title"] = nav(data, TITLE_TEXT)
         runs = nav(data, ["subtitle", "runs"])
--- a/ytmusicapi/parsers/search.py
+++ b/ytmusicapi/parsers/search.py
@@ -34,7 +34,6 @@ def parse_top_result(data, search_result_types):
             search_result["videoType"] = nav(on_tap, NAVIGATION_VIDEO_TYPE)
 
     if result_type in ["song", "video", "album"]:
-        on_tap = data.get("onTap")
         search_result["videoId"] = nav(data, ["onTap"] + WATCH_VIDEO_ID, True)
         search_result["videoType"] = nav(data, ["onTap"] + NAVIGATION_VIDEO_TYPE, True)
 
--- a/ytmusicapi/parsers/search.py
+++ b/ytmusicapi/parsers/search.py
@@ -34,8 +34,13 @@ def parse_top_result(data, search_result_types):
             search_result["videoType"] = nav(on_tap, NAVIGATION_VIDEO_TYPE)
 
     if result_type in ["song", "video", "album"]:
+        on_tap = data.get("onTap")
+        if on_tap:
+            search_result["videoId"] = nav(on_tap, WATCH_VIDEO_ID)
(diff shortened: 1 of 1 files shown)


AustinGitHub:
Tried to run the pyTests but it's very complicated. The oauth works fine, but the browser one just says paste the request headers from firefox and press enter, but what request headers? I am confused on that part. 

Also, can't you write a new script to run both of those functions and then move the file as well? Would make it a lot easier.

jcbirdwell:
> but what request headers?

Request headers come from a post request to a user specific endpoint. So with your network monitor open navigate from something like home -> library and use the post request associated with it.

ytmusicapi/parsers/search.py
```diff
@@ -34,8 +34,17 @@ def parse_top_result(data, search_result_types):
...
 
     if result_type in ["song", "video", "album"]:
+        on_tap = data.get("onTap")
+        if on_tap is not None:
+            if "watchEndpoint" in on_tap:
+                if nav(on_tap, WATCH_VIDEO_ID) is not None:
+                    search_result["videoId"] = nav(on_tap, WATCH_VIDEO_ID)
+            if "navigationEndpoint" in on_tap:
+                if nav(on_tap, NAVIGATION_VIDEO_TYPE) is not None:
+                    search_result["videoType"] = nav(on_tap, NAVIGATION_VIDEO_TYPE)
```
sigma67, line 44: Instead of triple-if, just navigate directly to the key you want to access using `nav`. Make sure to set `none_if_absent` (i.e. `nav(data, keys, True)`) - this way it'll just error out with `None` at the first missing key in the chain of `keys`

```suggestion
       search_result["videoId"] = nav(data, ["onTap"] + WATCH_VIDEO_ID, True)
       search_result["videoType"] = nav(data, ["onTap"] + NAVIGATION_VIDEO_TYPE, True)
```

ytmusicapi/parsers/search.py
```diff
@@ -34,8 +34,17 @@ def parse_top_result(data, search_result_types):
...
+
         search_result["title"] = nav(data, TITLE_TEXT)
-        runs = nav(data, ["subtitle", "runs"])[2:]
+        runs = nav(data, ["subtitle", "runs"])
```
sigma67, line 47: why this change?

ytmusicapi/parsers/search.py
```diff
@@ -125,7 +134,7 @@ def parse_search_result(data, search_result_types, result_type, category):
         search_result["duration"] = None
         search_result["year"] = None
         flex_item = get_flex_column_item(data, 1)
-        runs = flex_item["text"]["runs"][default_offset:]
+        runs = flex_item["text"]["runs"]
```
sigma67, line 137: why this change?

sigma67 (commented):
Thanks for the contribution, see comments.

I suggest we discuss the setup issue over in #520. I think we could improve `CONTRIBUTING.md` and I'm open towards a script if it's feasible. Could probably also simplify the `test.cfg` a bit. 

Request headers are documented here: https://ytmusicapi.readthedocs.io/en/stable/setup/browser.html

But I can see how you'd miss that if you never bothered with the old way of doing things since we have oauth now

ytmusicapi/parsers/search.py
AustinGitHub: Sure I can do that

ytmusicapi/parsers/search.py
AustinGitHub: Like the other, this was part of the issue too

With my change
image

With original change
image

I crossed check the Json files I am writing too and I see with removing this offset it has better chance of showing the artist name with Id over nothing.

ytmusicapi/parsers/search.py
AustinGitHub: I made this change, thank you!

ytmusicapi/parsers/search.py
AustinGitHub: If you check the before and after jsons I attached to the PR, it will also show this difference

ytmusicapi/parsers/search.py
```diff
@@ -34,8 +34,12 @@ def parse_top_result(data, search_result_types):
             search_result["videoType"] = nav(on_tap, NAVIGATION_VIDEO_TYPE)
 
     if result_type in ["song", "video", "album"]:
+        on_tap = data.get("onTap")
```
sigma67, line 37:
```suggestion
```

AustinGitHub:
Thanks!

Issue #522, opened 2024-01-14
extend get_album


jcbirdwell:
**Is your feature request related to a problem? Please describe.**
I ran into an issue where the albums listed under "other_versions" had the artist name as the album year. When I went to correct it I also found a bunch of other keys that weren't being pulled out of the response. 

**Describe the solution you'd like**
Add the missing keys to value returned from get_album

**Describe alternatives you've considered**
None

**Additional context**
The original album I ran into the issue with was Eminem Curtain Call: The Hits, browseId: MPREb_LQCAymzbaKJ, pid: OLAK5uy_lL6bP3tsOcU9Tu3zCG61my265UG4wH5wU

PR #523, opened 2024-01-14
get_album: parse more data from response + correct keys on other_vers…

jcbirdwell:
#522 + tests. I'll bet there's still more to be found, but I'm going cross eyed trying to read these response objects. This fixes the original issue of incorrect keys in "other_versions" and adds audio playlist id, artists with id and name, album type, removes year (I couldn't find it), and adds explicit to the main album requested in get_album.


ytmusicapi/parsers/browsing.py
```diff
@@ -54,10 +54,20 @@ def parse_content_list(results, parse_func, key=MTRIR):
...
         "title": nav(result, TITLE_TEXT),
-        "year": nav(result, SUBTITLE2, True),
+        "type": nav(result, SUBTITLE0),
+        "artists": [parse_id_name(x) for x in nav(result, ["subtitle", "runs"]) if "navigationEndpoint" in x],
         "browseId": nav(result, TITLE + NAVIGATION_BROWSE_ID),
+        # making previously unavailable keys snakecase
+        "audio_playlist_id": nav(result, THUMBNAIL_OVERLAY, True),
         "thumbnails": nav(result, THUMBNAIL_RENDERER),
-        "isExplicit": nav(result, SUBTITLE_BADGE_LABEL, True) is not None,
+        "explicit": nav(result, SUBTITLE_BADGE_LABEL, True) is not None,
```
AustinGitHub, line 63: question, why change form `isExplicit` to `explicit` this can break how people have their scripts written already using this plugin. I recommend keeping the same name.

jcbirdwell: fair

ytmusicapi/parsers/browsing.py
```diff
@@ -54,13 +54,23 @@ def parse_content_list(results, parse_func, key=MTRIR):
 def parse_album(result):
     return {
         "title": nav(result, TITLE_TEXT),
-        "year": nav(result, SUBTITLE2, True),
```
sigma67, line 57: why remove this?

jcbirdwell: YT may have moved something, on all the albums I've tried that nav route gives you the album's artist. I can't find year anywhere.

ytmusicapi/navigation.py
```diff
@@ -41,11 +43,13 @@
 TEXT_RUN_TEXT = TEXT_RUN + ["text"]
 SUBTITLE = ["subtitle"] + RUN_TEXT
 SUBTITLE_RUNS = ["subtitle", "runs"]
+SUBTITLE0 = SUBTITLE_RUNS + [0, "text"]
```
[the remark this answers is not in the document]
jcbirdwell: Nice catch

ytmusicapi/parsers/browsing.py
```diff
@@ -54,13 +54,23 @@ def parse_content_list(results, parse_func, key=MTRIR):
...
+        "artists": [parse_id_name(x) for x in nav(result, ["subtitle", "runs"]) if "navigationEndpoint" in x],
         "browseId": nav(result, TITLE + NAVIGATION_BROWSE_ID),
+        # making previously unavailable keys snakecase
+        "audioPlaylistId": nav(result, THUMBNAIL_OVERLAY, True),
         "thumbnails": nav(result, THUMBNAIL_RENDERER),
         "isExplicit": nav(result, SUBTITLE_BADGE_LABEL, True) is not None,
     }
 
 
+def parse_id_name(sub_run):
```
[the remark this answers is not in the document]
jcbirdwell: I intentionally left it generic as the same pattern appears outside of artists. Didn't commit any related changes in this PR, but just about anything with a title and browse id could be pulled with it. That being said, its more than likely a weird duplicate of some other function I didn't see.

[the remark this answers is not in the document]
sigma67: There's not yet another function like it, so that is fine. Functions with similar parsing: `parse_song_flat`, `parse_song_artist_runs`, `parse_song_runs`

`parse_song_runs` and `parse_song_artist_runs` are places where you could use this function as well. I suggest refactoring these functions to use it

`parse_song_flat` is an example where you'd parse an `album` name and id - which works differently. Therefore I'd at least include the word `artist` in the function name: `parse_artist_id_name` - as artist is the only context we know where it works

ytmusicapi/parsers/browsing.py
sigma67: This function has multiple usages (via `parse_mixed_content` (get_home, get_song_related), `parse_artist_contents` (get_artist, get_user)), please double-check that this is in fact returning `None` for all of them

jcbirdwell: Looks like get_home's return is the same as get_album()["other_versions"], so it's also incorrectly returning artist names as the year. 

get_song_related and get_artist do correctly return year with the old version of parse_album, but could still benefit from type and audioplaylistid returns of the PR version.

get_user appears to be incapable of actually returning albums despite having access to the function, so I'm not sure on that one unless you have an ID I could test against. 

I'll rework the PR to flip between artists and year return based on the context.

sigma67 (commented):
Hi @jcbirdwell , please don't take this the wrong way, but your last commit went too far beyond the scope of this PR and the code/intention isn't really clear to me (especially what you did in `parse_pl_song_artists` and `parse_playlist` - `parse_video` also seemed unfinished).

So to include the original fixes in the next release I'm reverting the most recent commit and finishing up. Feel free to resubmit changes from the last commit in a separate PR.

PR #527, opened 2024-01-20
Fix OAUTH_CUSTOM_FULL mode

MarvinSchenkel:
This PR adds the missing `AuthType.OAUTH_CUSTOM_FULL` mode when calculating the headers. We (Music Assistant) use our own web-based OAuth process to retrieve an acces token. In this scenario we also provide the other required headers to interact with Youtube Music.

### DIFF
--- a/ytmusicapi/ytmusic.py
+++ b/ytmusicapi/ytmusic.py
@@ -211,6 +211,11 @@ class YTMusicBase:
         if self.auth_type == AuthType.BROWSER:
             self._headers["authorization"] = get_authorization(self.sapisid + " " + self.origin)
 
+        # Override the headers with the auth / input_dict when using OAUTH_CUSTOM_FULL
+        # Full headers are provided by the downstream client in this scenario.
+        elif self.auth_type == AuthType.OAUTH_CUSTOM_FULL:
+            self._headers = self._input_dict
+
         elif self.auth_type in AuthType.oauth_types():
             self._headers["authorization"] = self._token.as_auth()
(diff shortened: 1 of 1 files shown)


sigma67 (commented):
Shouldn't this already be supported via 

https://github.com/sigma67/ytmusicapi/blob/main/ytmusicapi/ytmusic.py#L190-L191

If not, can we have a test as well that covers the difference and ensures it won't break again in the future?

MarvinSchenkel:
Good point! In that case I would need to make some changes to this line, as the token is None for full custom oauth: https://github.com/sigma67/ytmusicapi/blob/917019745723d6ffacc32d7eb8c2676e5debba0c/ytmusicapi/ytmusic.py#L215. This currently throws an error.

sigma67:
Yeah, is it enough to just not set `authorization`  `if self.auth_type == AuthType.OAUTH_CUSTOM_FULL:` ?

MarvinSchenkel:
> Yeah, is it enough to just not set `authorization` `if self.auth_type == AuthType.OAUTH_CUSTOM_FULL:` ?

Absolutely, fixed that

sigma67:
@MarvinSchenkel  please submit to `main` next time, `master` is no longer used

MarvinSchenkel:
> @MarvinSchenkel please submit to `main` next time, `master` is no longer used

Sorry I wasn't aware. My existing fork was still pointing to master.

sigma67:
I deleted the `master` branch in this repo so this can't happen anymore 👍

PR #530, opened 2024-01-21
Album data mining fixes

sigma67:


sigma67:
fyi @jcbirdwell here are your changes from the last commit. Feel free to open a separate PR again

Issue #539, opened 2024-02-10
KeyError 'runs' on get_library_playlists()


linsomniac:
**Describe the bug**
I have a user over in https://github.com/linsomniac/spotify_to_ytmusic/issues/31 that is reporting the following traceback:

```
Traceback (most recent call last):
  File "/usr/local/bin/s2yt_copy_all_playlists", line 8, in <module>
    sys.exit(copy_all_playlists())
  File "/usr/local/lib/python3.10/site-packages/spotify2ytmusic/cli.py", line 264, in copy_all_playlists
    backend.copy_all_playlists(
  File "/usr/local/lib/python3.10/site-packages/spotify2ytmusic/backend.py", line 441, in copy_all_playlists
    dst_pl_id = get_playlist_id_by_name(yt, pl_name)
  File "/usr/local/lib/python3.10/site-packages/spotify2ytmusic/backend.py", line 163, in get_playlist_id_by_name
    for pl in yt.get_library_playlists(limit=5000):
  File "/usr/local/lib/python3.10/site-packages/ytmusicapi/mixins/library.py", line 35, in get_library_playlists
    playlists = parse_content_list(results["items"][1:], parse_playlist)
  File "/usr/local/lib/python3.10/site-packages/ytmusicapi/parsers/browsing.py", line 48, in parse_content_list
    contents.append(parse_func(result[key]))
  File "/usr/local/lib/python3.10/site-packages/ytmusicapi/parsers/browsing.py", line 126, in parse_playlist
    "title": nav(data, TITLE_TEXT),
  File "/usr/local/lib/python3.10/site-packages/ytmusicapi/navigation.py", line 103, in nav
    raise err
  File "/usr/local/lib/python3.10/site-packages/ytmusicapi/navigation.py", line 97, in nav
    root = root[k]
KeyError: 'runs'
```

It looks like this comes from "ytmusicapi/parsers/browsing.py" in line 126, which reads: `"title": nav(data, TITLE_TEXT),`, and "TITLE_TEXT" appears to be `["title", "runs", 0, "text"]`.  The record it is on apparently does not have a "runs" key.  Perhaps that call to "nav()" should have the "none_if_absent=True" argument set?

I don't really understand the code in question so I'm reluctant to suggest a fix, especially as I don't have direct access to the data that is triggering this issue (it's reported by a user of my software, issue linked above).

**To Reproduce**
It seems to largely be dependent on the users ytmusic environment, as I am not able to reproduce this.

[...]

PR #540, opened 2024-02-10
nav() now displays some debugging information on KeyError

linsomniac:
There's a user reporting a KeyError in nav(), so I've changed it to print out the key and dict for future debugging.  Also, a bare Exception was used to capture a dictionary failure, removed that since my change checks for the existence of the key in the dictionary. If this is objectionable at all, you probably should change it to catching a KeyError rather than catching Exception.

This is to gather information for #539

### DIFF
--- a/ytmusicapi/navigation.py
+++ b/ytmusicapi/navigation.py
@@ -92,14 +92,14 @@ def nav(root: Dict, items: List[Any], none_if_absent: Literal[True] = True) -> O
 
 def nav(root: Dict, items: List[Any], none_if_absent: bool = False) -> Optional[Any]:
     """Access a nested object in root by item sequence."""
-    for k in items:
-        try:
+    try:
+        for k in items:
             root = root[k]
-        except (KeyError, IndexError) as e:
-            if none_if_absent:
-                return None
-            raise type(e)(f"Unable to find '{k}' using path {items!r} on {root!r}, exception: {e}")
-    return root
+        return root
+    except (KeyError, IndexError) as e:
+        if none_if_absent:
+            return None
+        raise type(e)(f"Unable to find '{k}' using path {items!r} on {root!r}, exception: {e}")
 
 
 def find_object_by_key(object_list, key, nested=None, is_key=False):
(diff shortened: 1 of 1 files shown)


sigma67:
Did you test this change?

`nav` has more than 250 usages across the project, it's not something to change as a drive-by

Simply running the `get_album` tests fails instantly - you didn't consider that `items` may contain integer indices

linsomniac:
I couldn't originally get the tests working, I believe this change addresses that problem.

sigma67:
Sorry, it still doesn't work.

```
TypeError: 'NoneType' object is not subscriptable
```

linsomniac:
Sorry for the delay in getting back to this, busy end of the week.  I missed that exception, thanks for pointing it out.  I think this modification resolves that issue.

One issue I'm running into that perhaps you can help with is that I'm getting a bunch of config-related test failures, and I didn't see that None error among those other ones.  The test instructions say to run: `cp tests/test.example.cfg tests/test.cfg`, but the repo doesn't seem to have a "test.example.cfg" file in it, or any "cfg" file AFAICT.

sigma67:
Sorry about the `test.cfg` issue, I accidentally deleted the file instead of renaming it due to the presence of a gitignore on `**.cfg`

sigma67:
What's the advantage of moving the try-except inside the loop? Isn't that even more overhead?

linsomniac:
Having it in the loop means it has access to the value of "k" to be able to log.  It could be moved outside the loop but would lose that data, but might still provide enough information to track down, given more effort.

sigma67:
But it's bad: https://docs.astral.sh/ruff/rules/try-except-in-loop/

And with `nav` it matters, because it's used a lot.

linsomniac:
```
[I] sean@seans-laptop ~/p/realgoansible (master)> python3 -m timeit 'for x in range(10):' '   y = x + 1'
2000000 loops, best of 5: 199 nsec per loop
[I] sean@seans-laptop ~/p/realgoansible (master)> python3 -m timeit 'for x in range(10):' '   try:' '      y = x + 1' '
except:' '      pass'
1000000 loops, best of 5: 202 nsec per loop
```

LSP is telling me that "k is unbound" if I move the try/except out of the loop, but if I run  a test I am getting the loop variable past the loop:

```python
try:
    for x in [1,2,3,4]:
        if x == 3:
            1 / 0
except Exception as e:
    print(f"Failed, {x}")
```

Results in: "Failed, 3".

So, I've moved it out of the loop.

PR #552, opened 2024-02-24
Fix parsing of album and views in parse_playlist_items

fischcode:
Parse album and views columns as a function of flex column count and is_album property to cater for varying layouts with three and four flex columns


fischcode:
@sigma67 Please note that the three commits are independent contributions an that there are no dependencies between them whatsoever

ytmusicapi/mixins/playlists.py
```diff
@@ -112,10 +112,12 @@ def get_playlist(
         if not own_playlist:
             header = response["header"]["musicDetailHeaderRenderer"]
             playlist["privacy"] = "PUBLIC"
+            playlist["own_playlist"] = False
```
sigma67, line 115: I would prefer to call this `owned`, much simpler.

Also you can just set the attribute outside the if and reuse the local variable `own_playlist`, much simpler

ytmusicapi/parsers/playlists.py
```diff
@@ -44,15 +44,13 @@ def parse_playlist_items(results, menu_entries: Optional[List[List]] = None, is_
...
 
-        album = parse_song_album(data, 2)
+        album = parse_song_album(data, flex_column_count-1) if not is_album else None
 
-        views = None
-        if album and album["id"] is None:
-            # views currently only present on albums and formatting is localization-dependent -> no parsing
-            if (views := (get_item_text(data, 2))) is not None:
-                album = None
+        views = get_item_text(data, 2) if flex_column_count == 4 or is_album else None
```
sigma67, line 53: > flex_column_count == 4 or is_album

which cases does this condition cover? `is_album` is obvious, but when does the `count==4` apply when it's not an album?

ytmusicapi/parsers/podcasts.py
```diff
@@ -103,6 +103,7 @@ def parse_episodes(results) -> List[Dict]:
     episodes = []
     for result in results:
         data = nav(result, ["musicMultiRowListItemRenderer"])
+        thumbnails = nav(data, THUMBNAILS)[:2]
```
sigma67, line 106: why are you only taking the first two items? just leave out the `[:2]`

sigma67 (commented):
Thanks for the contribution, looks good mostly. Some minor comments & please fix linter issues :)

ytmusicapi/parsers/playlists.py
fischcode: I am covering
- Album with three columns, the last of which is views/plays:
grafik
- Playlist with three columns, the last of which is album:
grafik
- Artist songs from Artist overview with four columns, the third of which is views/plays: and the fourth of which is album:
grafik
- Artist song details with four columns, the third of which is views/plays: and the fourth of which is album:
grafik

ytmusicapi/parsers/podcasts.py
fischcode: Some podcasts episodes appear to contain technical images in addition to the episode related images, e.g.
https://i.ytimg.com/vi/dPw_SXs1QOA/hq720.jpg:
Podcast 'Smosh Dungeons and Dragons', beginning of first episode:
`podcast = [{
        'musicMultiRowListItemRenderer': {
            'trackingParams': 'CHEQ3-AIGAAiEwjssYvPvceEAxWU0UIFHRKZBZk=',
            'thumbnail': {
                'musicThumbnailRenderer': {
                    'thumbnail': {
                        'thumbnails': [{
                                'url': 'https://i.ytimg.com/vi/f-vhvYvI0ok/hqdefault.jpg?sqp=-oaymwEWCOADEI4CIAQqCggAEOADGC0guwJIWg&rs=AMzJL3lYQ_uibdN_dDCw6PWruHiO_dolmQ',
                                'width': 480,
                                'height': 270
                            }, {
                                'url': 'https://i.ytimg.com/vi/f-vhvYvI0ok/hq720.jpg?sqp=-oaymwEKCKAGEMIDIABIWg&rs=AMzJL3k6ozV_-vU-izDtPVWu-lbKy5wgAw',
                                'width': 800,
                                'height': 450
                            }, {
                                'url': 'https://i.ytimg.com/vi/f-vhvYvI0ok/hq720.jpg',
                                'width': 1280,
                                'height': 720
                            }
                        ]
                    },
                    'thumbnailCrop': 'MUSIC_THUMBNAIL_CROP_UNSPECIFIED',
                    'thumbnailScale': 'MUSIC_THUMBNAIL_SCALE_ASPECT_FIT',
                    'trackingParams': 'CHsQhL8CIhMI7LGLz73HhAMVlNFCBR0SmQWZ'
                }
            },
`
Usually I display the last image in the list as it normally has the best resolution. But here that would result in the wrong image being displayed. So far, I haven't found episodes where such a technical image appeared before third position. Not having come up with a better idea so far, I therefore had decided to drop all thumbnails after position two.
Just now I realize that the third image's URL doesn't contain a question mark, so an alternative approach would be to filter for thumbnails with URLs containing one:
thumbnails = list(filter(lambda tn: '?' in tn['url'], nav(data, THUMBNAILS)))

ytmusicapi/mixins/playlists.py
fischcode: OK, changed

ytmusicapi/parsers/playlists.py
sigma67: Thanks for this overview, that explains it!

ytmusicapi/parsers/podcasts.py
sigma67: I realize that the same image is present in multiple resolutions. This is the case for other thumbnails as well.

`ytmusicapi` always returns all thumbnails - it's up to the user to choose the appropriate size. You never know the precise use case in advance, so can't be too judgy on the input data filtering.

fischcode:
Sorry, I don't understand: what is linter?

sigma67 (approved):

PR #554, opened 2024-02-27
add package-data so pip installs working translations

tytgatlieven:
When installing this package through pip localisation files are not found:

FileNotFoundError: [Errno 2] No translation file found for domain: 'base'

At:
  /usr/lib/python3.10/gettext.py(607): translation
  /usr/lib/python3/dist-packages/ytmusicapi/ytmusic.py(164): __init__

Enabling include-package-data fixes this problem

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,7 +31,7 @@ readme = {file = ["README.rst"]}
 [tool.setuptools_scm]
 
 [tool.setuptools]
-include-package-data=false
+include-package-data=true
 
 [tool.setuptools.package-data]
 "*" = ["**.rst", "**.py", "**.mo"]


sigma67:
I cannot confirm this, all `*.mo` files are included as expected. `*.po` and `*.pot` are not present as they are only for assembling `*.mo`.

What is your use case?

tytgatlieven:
First of all: You are right, I apologise.

I'm using audiotube kde neon. ytmusicapi is an apt package there. Apparently the locales are not correct in that package.

When I installed ytmusicapi using pip it did not work. When I installed it using pip and set the include-data-package to true it miraculously worked. I have no idea why.

However: ytmusicapi does work when I first uninstall the ytmusicapi apt package completely, and then installed the pip package. So some kind of conflict seems to be the source.

sigma67:
No worries, glad to know everything is working as expected :)

PR #563, opened 2024-03-13
Add "type" attribute to library channels

fischcode:
YTMusic Library Channels list not only channels in the narrow sense but also artists publishing podcasts and episodes (e.g. The Offspring). To distinguish the two types of items, the json code contains a page_type attribute: [get_library_channels response.json](https://github.com/sigma67/ytmusicapi/files/14583273/get_library_channels.response.json). This PR adds code to read the page_type attribute and, if present, add a "tyoe" item with a value of "channel" or "artist" to the artist dictionary object.


sigma67:
So they are the same page after all. I guess at some point it makes sense to merge both functions... Behind the scenes they're using the same parser anyway

PR #580, opened 2024-05-09
Gracefully retrieve uploaded songs with a blank duration (#578)

### LINKED ISSUE #578: `KeyError: 'fixedColumns'` in specific circumstance when retrieving uploads
It seems like Google made a minor change with regard to uploading songs that is affecting `get_library_upload_songs()`

Now when you upload a new song via the web page, after the song finishes processing, it may first appear in your library without a duration:

image

Then after about 10 seconds or so, when you refresh it will have the duration:

image

During the brief period of time when there is not yet a duration, `get_library_upload_songs()` throws a KeyError for the `fixedColumns` attribute, since that entire attribute will be missing:
```
  File "/home/apastel/dev/ytmusic-deleter/test.py", line 5, in <module>
    response = yt_auth.get_library_upload_songs()
  File "/home/apastel/dev/ytmusic-deleter/.venv/lib/python3.10/site-packages/ytmusicapi/mixins/uploads.py", line 59, in get_library_upload_songs
    songs = parse_uploaded_items(results["contents"])
  File "/home/apastel/dev/ytmusic-deleter/.venv/lib/python3.10/site-packages/ytmusicapi/parsers/uploads.py", line 33, in parse_uploaded_items
    duration = get_fixed_column_item(data, 0)["text"]["runs"][0]["text"]
  File "/home/apastel/dev/ytmusic-deleter/.venv/lib/python3.10/site-packages/ytmusicapi/parsers/_utils.py", line 48, in get_fixed_column_item
    "text" not in item["fixedColumns"][index]["musicResponsiveListItemFixedColumnRenderer"]
KeyError: 'fixedColumns'
```

I feel like this is a new change from Google because I'm suddenly experiencing it in unit tests and one user of my app reported it today. It probably reduces the upload processing time on their part by letting the duration get filled in lazily afterwards. I was able to reproduce it outside my app in just a basic test file, and I ran this repeatedly after uploading a song.
```test.py

yt_auth = YTMusic("./tests/resources/oauth.json")
response = yt_auth.get_library_upload_songs()
print(response)
```

So it probably amounts to an attribute check and maybe we just don't include the duration in the results (or `None`) if that happens?

[without_fixedColumns.json](https://github.com/sigma67/ytmusicapi/files/15214851/without_fixedColumns.json)

apastel:
Fixes #578 

This prevents a possible KeyError or a int-parsing error when an uploaded song has a duration that is not a valid value like "3:31".

Per-recent changes by Google, an uploaded song can have a duration of `' '` or simply not have a `fixedColumns` object at all. This is most commonly seen immediately after uploading a song but can be seen in other cases as well.

I added a unit test that exercises this. It uploads a song and then starts calling `get_library_upload_songs()` in a loop, waiting for the song to be available. Without my duration fix, this encounters a parsing error. With the fix, it succeeds.

### DIFF
--- a/tests/mixins/test_uploads.py
+++ b/tests/mixins/test_uploads.py
@@ -1,6 +1,5 @@
 import tempfile
 import time
-from pathlib import Path
 
 import pytest
 
--- a/tests/mixins/test_uploads.py
+++ b/tests/mixins/test_uploads.py
@@ -63,9 +63,7 @@ class TestUploads:
             # Need to wait for song to be fully deleted
             time.sleep(10)
             # Now re-upload
-            upload_response = yt_auth.upload_song(
-                get_resource(config["uploads"]["file"])
-            )
+            upload_response = yt_auth.upload_song(get_resource(config["uploads"]["file"]))
 
         assert (
             upload_response == "STATUS_SUCCEEDED" or upload_response.status_code == 200
--- a/tests/mixins/test_uploads.py
+++ b/tests/mixins/test_uploads.py
@@ -1,5 +1,6 @@
 import tempfile
 import time
+from pathlib import Path
 
 import pytest
 
@@ -54,7 +55,7 @@ class TestUploads:
         upload_response = yt_auth.upload_song(get_resource(config["uploads"]["file"]))
         if not isinstance(upload_response, str) and upload_response.status_code == 409:
             # Song is already in uploads. Delete it and re-upload
-            songs = yt_auth.get_library_upload_songs()
+            songs = yt_auth.get_library_upload_songs(limit=None, order="recently_added")
             delete_response = None
             for song in songs:
                 if song.get("title") in config["uploads"]["file"]:
@@ -70,10 +71,10 @@ class TestUploads:
         ), f"Song failed to upload {upload_response}"
 
         # Wait for upload to finish processing and verify it can be retrieved
-        retries_remaining = 15
+        retries_remaining = 5
         while retries_remaining:
-            time.sleep(3)
-            songs = yt_auth.get_library_upload_songs()
+            time.sleep(5)
+            songs = yt_auth.get_library_upload_songs(limit=None, order="recently_added")
             for song in songs:
                 if song.get("title") in config["uploads"]["file"]:
                     # Uploaded song found
--- a/tests/mixins/test_uploads.py
+++ b/tests/mixins/test_uploads.py
@@ -1,4 +1,5 @@
 import tempfile
+import time
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


tests/mixins/test_uploads.py
```diff
@@ -47,6 +48,26 @@ def test_upload_song(self, config, yt_auth):
         response = yt_auth.upload_song(get_resource(config["uploads"]["file"]))
         assert response.status_code == 409
 
+    def test_upload_song_and_verify(self, config, yt_auth):
+        """Upload a song and verify it can be retrieved after it finishes processing."""
+        response = yt_auth.upload_song(get_resource(config["uploads"]["file"]))
+        if not isinstance(response, str) and response.status_code == 409:
+            raise AssertionError(f"{config['uploads']['file']!r} was already uploaded. Delete it and re-run.")
```
sigma67, line 55: This is not something that is reproducible. This test would always fail on CI (see `test_upload_song` - it asserts precisely that the return code is 409).

We could do something like the playlist end2end test: delete the uploaded song if it exists (which it will because `test_upload_song` runs before), upload, run your test below.

This would be a great addition because right now we don't actually run the deletion test below - for all I know it could have stopped working two years ago

sigma67 (commented):
Thanks for your contribution and work towards improving the reliability of the uploads functions :)

tests/mixins/test_uploads.py
```diff
@@ -47,6 +49,41 @@ def test_upload_song(self, config, yt_auth):
...
+        if not isinstance(upload_response, str) and upload_response.status_code == 409:
+            # Song is already in uploads. Delete it and re-upload
+            songs = yt_auth.get_library_upload_songs()
+            delete_response = None
+            for song in songs:
+                if song.get("title") in config["uploads"]["file"]:
+                    delete_response = yt_auth.delete_upload_entity(song["entityId"])
+            assert delete_response == "STATUS_SUCCEEDED"
+            # Need to wait for song to be fully deleted
+            time.sleep(10)
```
apastel, line 64: I didn't like adding a hard sleep of 10 seconds here but it turns out that even though the delete response was "success" and a subsequent call to `get_library_uploads()` would yield an empty list, the next `upload_song()` call was still resulting in a 409 Conflict. 10 seconds seems to consistently be enough to ensure that it's fully deleted.

tests/mixins/test_uploads.py
apastel: Ok, so now the test seems very robust. I can run it repeatedly without waiting in between runs and it passes consistently, even if the other `test_upload_song` was just ran.

tests/mixins/test_uploads.py
sigma67: Fine with me, there are some similar issue with the playlist end2end returning conflicts. Better than a flaky test

tests/mixins/test_uploads.py
sigma67: In general it looks good to me, but it's failing for me:

```python
yt_auth = <ytmusicapi.ytmusic.YTMusic object at 0x000001B473EACC90>

def test_upload_song_and_verify(self, config, yt_auth: YTMusic):
    """Upload a song and verify it can be retrieved after it finishes processing."""
    upload_response = yt_auth.upload_song(get_resource(config["uploads"]["file"]))
    if not isinstance(upload_response, str) and upload_response.status_code == 409:
        # Song is already in uploads. Delete it and re-upload
        songs = yt_auth.get_library_upload_songs()
        delete_response = None
        for song in songs:
            if song.get("title") in config["uploads"]["file"]:
                delete_response = yt_auth.delete_upload_entity(song["entityId"])
>           assert delete_response == "STATUS_SUCCEEDED"
E           AssertionError: assert None == 'STATUS_SUCCEEDED'

tests\mixins\test_uploads.py:62: AssertionError
```

The other upload test passes.

sigma67: I see, it's because I have a lot of uploaded songs and you didn't set `limit=None`.

Issue #597, opened 2024-05-31
get_playlist fails if some track has invalid duration_seconds


adwait-thattey:
**Describe the bug**
Some tracks in Yt music can have invalid entry for field `duration_seconds`. I typically saw this with unavailable tracks (that were live radios, but now removed) but maybe others can have similar issue. See attached sample json for the track below. 

Due to this, if a playlist contains such tracks and we call `get_playlist` , the internal helper function [sum_total_duration](https://github.com/sigma67/ytmusicapi/blob/33098ba69eac6c3e5fe88ed502cd3da000996536/ytmusicapi/helpers.py#L71) fails because `duration_seconds` is not an int. 
See screenshot for failure below

2 sample tracks with invalid duration_seconds entry:
![image](https://github.com/sigma67/ytmusicapi/assets/33225124/59b2d1d1-3ca7-400e-b237-8cdd03ceb1a6)
![image](https://github.com/sigma67/ytmusicapi/assets/33225124/8202c6c0-9cdf-4699-a344-9f0af15c4372)

Sample track in my playlist
![image](https://github.com/sigma67/ytmusicapi/assets/33225124/9af6ef55-5df1-410a-9902-8e732b1b123b)

Error when get_playlist is called
![image](https://github.com/sigma67/ytmusicapi/assets/33225124/5a260a52-2963-4384-bc4f-8fca0799a52e)

**To Reproduce**
1. Have a playlist with a track that has invalid duration_seconds
2. call get_playlist on the playlist
3. See error

**Fix**
I have fixed this locally. Submitting a PR

PR #598, opened 2024-05-31
fix sum_total_duration failing if duration_seconds is not an int

adwait-thattey:
Fix for issue #597

### DIFF
--- a/ytmusicapi/helpers.py
+++ b/ytmusicapi/helpers.py
@@ -71,4 +71,7 @@ def to_int(string):
 def sum_total_duration(item):
     if "tracks" not in item:
         return 0
-    return sum([track["duration_seconds"] if "duration_seconds" in track else 0 for track in item["tracks"]])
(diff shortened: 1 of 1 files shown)


sigma67:
Hi @adwait-thattey , thanks for the contribution. I'd prefer to fix the issue at the source, this is more of a bandaid fix. 

Can you link a public playlist where the issue occurs?

adwait-thattey:
@sigma67 I dont have any public playlist, but I suppose you could try adding the following radio tracks to any playlist from YouTube then open the playlist in YTMusic to get the same effect

https://www.youtube.com/watch?v=36YnV9STBqc&list=PLYznTKpI2I8LiIcp80isOHqljtze6fbeS&index=21

https://www.youtube.com/watch?v=tNkZsRW7h2c&list=PLYznTKpI2I8LiIcp80isOHqljtze6fbeS&index=60

Issue #601, opened 2024-06-16
Edge case in edit_playlist: moving a title to the end of the playlist


AltoRetrato:
Hello, and thanks for the ytmusicapi! :D

The `edit_playlist()` method currently includes `ACTION_MOVE_VIDEO_BEFORE`, `setVideoId`, and `movedSetVideoIdSuccessor` in `body["actions"]` whenever the `moveItem` argument is defined. This mirrors the behavior of the YouTube Music client when a title is moved to a position above another one in a playlist.

However, **when moving a title to the end of a playlist**, the YouTube Music client performs a similar action but omits the `movedSetVideoIdSuccessor` field.

In order to allow ytmusicapi to do this as well, I suggest changing the `edit_playlist()` function from:
```python
        moveItem: Optional[Tuple[str, str]] = None,
[...]
        if moveItem:
            actions.append(
                {
                    "action": "ACTION_MOVE_VIDEO_BEFORE",
                    "setVideoId": moveItem[0],
                    "movedSetVideoIdSuccessor": moveItem[1],
                }
            )
```
to:
```python
        moveItem: Optional[Union[str, Tuple[str, str]]] = None,
[...]
        if moveItem:
            action = {
                        "action": "ACTION_MOVE_VIDEO_BEFORE",
                        "setVideoId": moveItem if isinstance(moveItem, str) else moveItem[0],
                     }
            if isinstance(moveItem, tuple) and len(moveItem) > 1:
                action["movedSetVideoIdSuccessor"] = moveItem[1]
            actions.append(action)
```

This way, `moveItem` can be a `Tuple[str,str]` (as it is today), as well as a single `str` or a `Tuple[str,]`, allowing to move a title to the end of the playlist.

sigma67:
Interesting that you're the first one to stumble upon this, I haven't looked at this function in ages. Will have a closer look the next few days. 

You should definitely be able to move an item to the end, if not I'd consider that a bug.

sigma67:
Contribution welcome.

AltoRetrato:
> You should definitely be able to move an item to the end, if not I'd consider that a bug.

If I understand this correctly, without this change, you can only move an item `A` above another item, so it can never be moved directly to the bottom of a playlist. You can do it with two steps, though - by first moving item `A` to the position above the last item in the list (`L`), and then moving `L` above `A`.

You can create a playlist with a few items and move one to the bottom to see the "action" parameters in the Javascript Console.

Anyway, I did pull request #602.

PR #602, opened 2024-06-16
edit_playlist(): Allow moving title to bottom of playlist

AltoRetrato:
Patch to allow `edit_playlist()` to move a title to the end of a playlist. See #601.

### DIFF
--- a/.github/workflows/coverage-pr.yml
+++ b/.github/workflows/coverage-pr.yml
@@ -1,12 +1,6 @@
 name: Code coverage
 
 on:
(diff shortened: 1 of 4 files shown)


sigma67:
Can you please add a test that asserts that you are now able to move to the end of a playlist?

You could modify/extend this test: https://github.com/sigma67/ytmusicapi/blob/33098ba69eac6c3e5fe88ed502cd3da000996536/tests/mixins/test_playlists.py#L67

Issue #608, opened 2024-06-21
`get_playlist` does not work for all languages


JohnHKoh:
**Describe the bug**
I receive the following errors when trying to run `get_playlist` with the Korean ("`ko`") language using v1.7.3:

```
Traceback (most recent call last):
  File "C:\Users\asdf\AppData\Local\Programs\Python\Python311\Lib\site-packages\ytmusicapi\helpers.py", line 64, in to_int
    int_value = locale.atoi(number_string)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\asdf\AppData\Local\Programs\Python\Python311\Lib\locale.py", line 342, in atoi
    return int(delocalize(string))
           ^^^^^^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: ''
```
and

```
  File "C:\Users\asdf\Source\MelonYTMPlaylist\src\playlist_updater.py", line 117, in get_playlist
    return self.ytmusic.get_playlist(self.playlistId, None)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\asdf\AppData\Local\Programs\Python\Python311\Lib\site-packages\ytmusicapi\mixins\playlists.py", line 111, in get_playlist
    return self._parse_new_playlist_format(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\asdf\AppData\Local\Programs\Python\Python311\Lib\site-packages\ytmusicapi\mixins\playlists.py", line 215, in _parse_new_playlist_format
    song_count = to_int(song_count[0]) if len(song_count) > 1 else 0
                 ^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\asdf\AppData\Local\Programs\Python\Python311\Lib\site-packages\ytmusicapi\helpers.py", line 67, in to_int
    int_value = int(number_string)
                ^^^^^^^^^^^^^^^^^^
ValueError: invalid literal for int() with base 10: ''
```

**To Reproduce**
1. Set `YTMusic` language parameter to Korean (`language="ko"`)
2. Run `ytmusic.get_playlist(playlistId)`
3. See error

**Additional context**
https://github.com/sigma67/ytmusicapi/blob/11823d638719d76456ec58a6d7b76133368e5f1a/ytmusicapi/mixins/playlists.py#L214-L215
[...]

PR #609, opened 2024-06-21
Extract song count from text

JohnHKoh:
Use RegEx search to search for track count integer instead of using indexing. If the number cannot be extracted, return 0 as a fallback. I never hit the error in parsers, but seeing as the logic is the same, I went ahead and changed it there too.

Addresses https://github.com/sigma67/ytmusicapi/issues/608

### DIFF
--- a/ytmusicapi/parsers/playlists.py
+++ b/ytmusicapi/parsers/playlists.py
@@ -38,8 +38,10 @@ def parse_playlist_header(response: Dict) -> Dict[str, Any]:
         playlist["duration"] = (
             None if not has_duration else second_subtitle_runs[has_views + has_duration]["text"]
         )
-        song_count = second_subtitle_runs[has_views + 0]["text"].split(" ")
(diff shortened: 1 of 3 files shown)


ytmusicapi/parsers/playlists.py
```diff
@@ -38,8 +38,10 @@ def parse_playlist_header(response: Dict) -> Dict[str, Any]:
         playlist["duration"] = (
             None if not has_duration else second_subtitle_runs[has_views + has_duration]["text"]
         )
-        song_count = second_subtitle_runs[has_views + 0]["text"].split(" ")
-        song_count = to_int(song_count[0]) if len(song_count) > 1 else 0
+        song_count_text = second_subtitle_runs[has_views + 0]["text"]
+        song_count_search = re.search(r"\d+", song_count_text)
```
sigma67, line 42: Thanks for the contribution. Could you please add a test to check that it works for all supported languages?

Not sure what oddities we might be missing here.

Issue #622, opened 2024-07-14
`get_playlist` returns `None` for track title


abhishekmj303:
**Describe the bug**
For some playlists in YT Music Featured playlists like: [Indian Indie Essentials](https://music.youtube.com/playlist?list=RDCLAK5uy_kwmNQ6C3BXrVOcih6fGljxeeekGrzjTtE) and [Indie Rising](https://music.youtube.com/playlist?list=RDCLAK5uy_n17q7_2dwfDqWckpccDyTTkZ-g03jXuII) returns `None` for some of the track titles.

For example:
```json
{
    "videoId": "bSAlE_WgHxY",
    "title": null,
    "artists": [
        {
            "name": "Kanishk Seth",
            "id": "UC0PQFdpMlhl5TFYaUFy64yw"
        }
    ],
    "album": null,
    "likeStatus": "INDIFFERENT",
    "inLibrary": null,
    "thumbnails": [
        {
            "url": "https://i.ytimg.com/vi/bSAlE_WgHxY/sddefault.jpg?sqp=-oaymwEWCJADEOEBIAQqCghqEJQEGHgg6AJIWg&rs=AMzJL3nyvyq_o4DzAds4K5E5lFtH7Tynyg",
            "width": 400,
            "height": 225
        }
    ],
    "isAvailable": true,
    "isExplicit": false,
    "videoType": null,
    "views": null,
    "duration": "3:23",
    "duration_seconds": 203
}
```

Complete response: https://pastebin.com/rWCpb7uT

**To Reproduce**
1. Create a `YTMusic()` object
2. Call `get_playlist(playlist_id, limit=None)`
3. Error in getting track's title

**Additional context**
ytmusicapi version: 1.7.5
Python version: 3.11.9

Edit:
Just tested with other versions, and version 1.7.3 seems to work fine.

czifumasa:
I think it's the result of my changes in (#612). The playlist track from your example is not a song, but a podcast episode:
https://music.youtube.com/podcast/bSAlE_WgHxY

It needs another condition in a new logic. I created PR (#623) with a quick fix. After that, it should work the same as in 1.7.3.

sigma67:
@abhishekmj303 please verify that the fix works for you

abhishekmj303:
@czifumasa @sigma67 Yeah, it seems to be working now! Thank You.

PR #612, opened 2024-06-30
get_playlist: title, artists and album are correctly resolved in collaborative playlists (#603)

czifumasa:
Fixed invalid playlist item info for collaborative playlists from #603 issue.
Instead of hardcoded indexes for flexColumns, now they are resolved using their navigationEndpoint with following logic:

- `watchEndpoint` -> title
- `browseEndpoint` with `pageType == MUSIC_PAGE_TYPE_ALBUM` -> album
- `browseEndpoint` with `pageType == MUSIC_PAGE_TYPE_ARTIST` (Regular Artists) or `pageType == MUSIC_PAGE_TYPE_UNKNOWN` (Artists for uploaded songs)-> artists
- `browseEndpoint` with `pageType == MUSIC_PAGE_TYPE_USER_CHANNEL` -> user channel

Additional checks:
- sometimes song does not have navigable artist (example 4th track in https://music.youtube.com/playlist?list=OLAK5uy_mfa2N_1Ro0GZKmcrHLR_1VqSr30efpwTc). Then first unrecognized column is treated as artist.
- for non-song videos the last user channel is treated as artist
    - In regular playlists non-song videos have single user channel column for the channel that uploaded video,
    - In collaborative playlists non-song videos have two user channel columns. First column for the channel that added video to the playlist. The song column for the channel that uploaded video.

Unavailable songs do not have navigationEndpoints so hardcoded indexes are still used. Collaborative user is not listed for unavailable song, so it should not be a problem.

PR #623, opened 2024-07-14
get_playlist: fix title for podcast episodes (#622)

czifumasa:
Quick fix for get_playlist method. Some playlist items had None as a title. (#622)

After refactor in (#612),  title is resolved using `watchEndpoint`. Podcast episodes do not have one. They have `browseEndpoint` with `pageType = MUSIC_PAGE_TYPE_NON_MUSIC_AUDIO_TRACK_PAGE` instead.


sigma67:
Hi @czifumasa thanks for your persistence here.

On the weekend I also played around with `get_playlist` and found that for `suggestions` only, the author is not parsed correctly using your approach. Maybe you can have a look if you have time.

PR #637, opened 2024-08-06
Very small additions to documentation

indierodo:
While I was reading the docs to append video items, I got confused by the setVideoId parameter. I added the missing field to the get_playlists_id example and copied the definition to the edit playlist documentation.

I hope this is ok, it's a very small change, but I think it might help others. I really appreciate the time you take by looking at this.


indierodo:
I apologize, I did not realize that this fields only appear when you query a playlist you own.

iXenonN:
how you create the oauth.json file when i try to create with ytmusicapi command in cmd or powershell it says 'ytmusicapi' is not recognized as an internal or external command, operable program or batch file.

PR #651, opened 2024-09-28
fix(playlists): handle missing regex group in trackCount

### LINKED ISSUE #650: Incorrect track count for playlists with 1000+ songs
Hello all,

This is an issue with the regex in the playlist mixin. It causes only the first group to be returned. For example if the playlist had `3647` songs it would return `3` due to how `re.search(r"\d+", song_count_text)` is used.

**To Reproduce**
1. Find a playlist with 1000+ songs
2. Use the `YTMusic.get_playlist()` function
3. Print the `trackCount` object.
4. It should return only the first digit

**Additional context**
```python

# Initialize YouTube Music API
ytmusic = YTMusic()

PLAYLIST_ID = "XXXXXX"

# Call get_playlist
playlist = ytmusic.get_playlist(PLAYLIST_ID)

#Print trackCount to the console
print(playlist["trackCount"])
```

RealCalumPlays:
Fixes #650

### DIFF
--- a/tests/mixins/test_playlists.py
+++ b/tests/mixins/test_playlists.py
@@ -41,43 +41,29 @@ class TestPlaylists:
                 if track["videoType"] == "MUSIC_VIDEO_TYPE_ATV":
                     assert isinstance(track["album"]["name"], str) and track["album"]["name"]
 
-    def test_get_playlist_foreign(self, yt, yt_auth, yt_oauth):
+    def test_get_playlist_foreign(self, yt_empty, yt_oauth, sample_playlist):
         with pytest.raises(Exception):
-            yt.get_playlist("PLABC")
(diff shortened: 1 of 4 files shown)


sigma67:
Hi @RealCalumPlays , thanks for your contribution. Could you please add a test that covers this case? Best to use a public playlist with 1000+ entries.

RealCalumPlays:
Yup can do that for ya. Sorry for the late response. Do you want another PR when I have added it?

sigma67:
@RealCalumPlays Thanks! No, just add it here. Please add it next to the existing tests for `get_playlist`. Since it's a public playlist you shouldn't have to bother with the configuration setup I hope.

RealCalumPlays:
Okay I have just added it to `TestPlaylists::test_get_playlist_foreign_new_format` and `TestPlaylists::test_get_playlist_foreign` (I think thats the place you wanted me to do it?) I have also updated the first playlist ID used in the first check in those functions as they do not work anymore causing the tests to always fail.

sigma67:
@RealCalumPlays Thanks! Quality playlist choice. I'll have a look

PR #657, opened 2024-10-05
fix #656: attempt to find runs in subtitle if not in title

### LINKED ISSUE #656: "Unable to find 'runs' using path ['title', 'runs', 0, 'text'] on {}
**Describe the bug**

Under some circumastances, YT returns a playlist where "runs" is under "subtitle" and not "title". 

**To Reproduce**

```
>>> from ytmusicapi.ytmusic import YTMusic
>>> y = YTMusic(None)
>>> y.get_playlist("PLZ6Ih9wLHQ2Hm2d3Cb0iV48Z2hQjGRyNz", limit=None)
```

The error I get is

```
  File "/Users/fritz/work/ytm2spt/venv/lib/python3.12/site-packages/ytmusicapi/navigation.py", line 106, in nav
    root = root[k]
           ~~~~^^^
KeyError: 'runs'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/Users/fritz/work/ytm2spt/venv/lib/python3.12/site-packages/ytmusicapi/mixins/playlists.py", line 111, in get_playlist
    return self._parse_new_playlist_format(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fritz/work/ytm2spt/venv/lib/python3.12/site-packages/ytmusicapi/mixins/playlists.py", line 201, in _parse_new_playlist_format
    playlist["title"] = nav(header, TITLE_TEXT)
                        ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/fritz/work/ytm2spt/venv/lib/python3.12/site-packages/ytmusicapi/navigation.py", line 110, in nav
    raise type(e)(f"Unable to find '{k}' using path {items!r} on {root!r}, exception: {e}")
KeyError: "Unable to find 'runs' using path ['title', 'runs', 0, 'text'] on {}, exception: 'runs'"
```

**Additional context**

It could be that YT messed up on their side - for example the playlists where that happens are not listed in my user account but they are there if I access them by direct playlist id. They used to be there, and something is odd. Either way, it _does_ happen, and ytmusicapi should be able to deal with it.

It's an easy fix and I will submit a PR

gotofritz:
Sometimes the object 'runs' is not to be found within title, but within
subtitle. This PR makes ytmusicapi look in both places.

Closes #656

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,12 +34,6 @@ def fixture_sample_video() -> str:
     return "hpSrLjc5SMs"
 
 
-@pytest.fixture(name="sample_playlist")
-def fixture_sample_playlist() -> str:
-    """very large playlist"""
-    return "PL6bPxvf5dW5clc3y9wAoslzqUrmkZ5c-u"
-
-
 @pytest.fixture(name="browser_filepath")
 def fixture_browser_filepath(config) -> str:
     return get_resource(config["auth"]["browser_file"])
--- a/tests/mixins/test_playlists.py
+++ b/tests/mixins/test_playlists.py
@@ -42,31 +42,32 @@ class TestPlaylists:
                 if track["videoType"] == "MUSIC_VIDEO_TYPE_ATV":
                     assert isinstance(track["album"]["name"], str) and track["album"]["name"]
 
-    def test_get_playlist_foreign(self, yt_empty, yt_oauth, sample_playlist):
-        with pytest.raises(Exception):
-            yt_empty.get_playlist("PLABC")
-        playlist = yt_empty.get_playlist(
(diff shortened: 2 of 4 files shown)


sigma67:
@gotofritz I would like to merge the fix but as you can tell from the pipeline I cannot reproduce..  In the second commit I removed your fix and the result is the same, no crash.

It seems this bug is only present when you own the playlist?

Here is a screenshot what it looks like for me, logged in. You could also test if the issue also occurs with a non-authenticated account for you.

gotofritz:
Yeah it's not happening any more, it must have been a short lived glitch. The playlists have titles again and they now appear in my list. FYI, my account was suspended out of the blue for three hours and then restored, not sure what happened there but this must have been part of it. Thanks for your time, not sure what you want to do with this PR now, up to you.

PR #661, opened 2024-10-10
Fix duration parsing for very long videos

### LINKED ISSUE #660: `parse_duration` fails for very long videos
**Describe the bug**
`get_playlist` fails when the playlist contains very long videos like [this](https://www.youtube.com/watch?v=hHW1oY26kxQ&list=PLfuGZPdCf-K_65ebc5u1yc5WonSf5HOXk&index=264&pp=iAQB8AUB). These are basically streams converted to videos.

The above mentioned duration is: `13,165:23:44`

**To Reproduce**
1. Call `get_playlist("PLfuGZPdCf-K_65ebc5u1yc5WonSf5HOXk")`
2. Get error:
```
/lib/python3.11/site-packages/ytmusicapi/parsers/_utils.py", line 71, in <genexpr>
    seconds = sum(multiplier * int(time) for multiplier, time in mapped_increments)
                               ^^^^^^^^^
ValueError: invalid literal for int() with base 10: '13,165'
```

abhishekmj303:
Fixes #660

### DIFF
--- a/ytmusicapi/parsers/_utils.py
+++ b/ytmusicapi/parsers/_utils.py
@@ -69,8 +69,8 @@ def parse_duration(duration):
         return duration
     duration_split = duration.strip().split(":")
     for d in duration_split:
-        if not d.isdigit(): # For e.g: "2,343"
-            return
+        if not d.isdigit():  # For e.g: "2,343"
+            return None
     mapped_increments = zip([1, 60, 3600], reversed(duration_split))
     seconds = sum(multiplier * int(time) for multiplier, time in mapped_increments)
     return seconds
--- a/ytmusicapi/parsers/_utils.py
+++ b/ytmusicapi/parsers/_utils.py
@@ -67,6 +67,7 @@ def parse_duration(duration):
     # duration may be falsy or a single space: ' '
     if not duration or not duration.strip():
         return duration
+    duration = duration.replace(",", "") # may be more than thousand hours
     mapped_increments = zip([1, 60, 3600], reversed(duration.split(":")))
     seconds = sum(multiplier * int(time) for multiplier, time in mapped_increments)
     return seconds
--- a/ytmusicapi/parsers/_utils.py
+++ b/ytmusicapi/parsers/_utils.py
@@ -67,8 +67,11 @@ def parse_duration(duration):
     # duration may be falsy or a single space: ' '
     if not duration or not duration.strip():
         return duration
-    duration = duration.replace(",", "") # may be more than thousand hours
-    mapped_increments = zip([1, 60, 3600], reversed(duration.split(":")))
+    duration_split = duration.strip().split(":")
+    for d in duration_split:
+        if not d.isdigit(): # For e.g: "2,343"
+            return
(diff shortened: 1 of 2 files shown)


sigma67:
So I'm not sure what the correct behavior is here...

Afaik from sources on the internet, YT's max video duration is 12h. 

Since the video you linked is not playable either I would assume that it's simply an invalid entry. Not sure we'd even want to parse the duration in that case. 

Granted we should not crash but return `None` as `duration` instead.

sigma67:
The fix you provided here is also locale dependent and would break for German localization

`duration: 13.165:23:44`

abhishekmj303:
> The fix you provided here is also locale dependent and would break for German localization
> 
> `duration: 13.165:23:44`

Good point, I did not think of this. If the locale is known then we can use the function `locale.atoi`. [Reference](https://stackoverflow.com/a/1779324)

But if this is not possible then this the way:
> Granted we should not crash but return None as duration instead.

sigma67:
Are there any videos with 1k+ hours that are actually playable? If yes I'd have a go with atoi. If no, don't bother and just return None

sigma67:
fyi there is already a `to_int` function that uses `atoi` 

https://github.com/sigma67/ytmusicapi/blob/2b5d19abb1820cb719b5abc817370aee90b9c5f5/ytmusicapi/helpers.py#L60

abhishekmj303:
> Are there any videos with 1k+ hours that are actually playable?

I think they are not playable. The current longest playable video on youtube I was able to find is about 300hrs ([https://youtu.be/kj4LxVhx2-A](https://youtu.be/kj4LxVhx2-A)). So just returning None should be fine.

sigma67:
@abhishekmj303 ok yes, let's do that.

I guess you could check if the result of `duration.split(":")` contains any non-digit characters and then fall back to None

abhishekmj303:
@sigma67 Not sure what tests and where I need to add them. Can you help me here?

ytmusicapi/parsers/_utils.py
```diff
@@ -67,7 +67,11 @@ def parse_duration(duration):
     # duration may be falsy or a single space: ' '
     if not duration or not duration.strip():
         return duration
-    mapped_increments = zip([1, 60, 3600], reversed(duration.split(":")))
+    duration_split = duration.strip().split(":")
+    for d in duration_split:
+        if not d.isdigit(): # For e.g: "2,343"
+            return
```
sigma67, line 73: Did you mean to return `None` ?

sigma67 (commented):
Please add a test with a playlist that contains an applicable song with >1k hours

ytmusicapi/parsers/_utils.py
abhishekmj303: Both `return` and `return None` are the same in python

sigma67: True, but being explicit is preferred here

https://stackoverflow.com/a/15300671

sigma67:
@abhishekmj303 Thanks! Everything looks good now

PR #662, opened 2024-10-10
Added Lyrics w. Timestamps

heinrich26:
Added `get_lyrics_with_timestamps` to get lyrics with timestamps. The Method doesn't try to parse the response as normal lyrics, if no lyrics with timestamps are returned. (could be changed to do so tho, the format is the same)


sigma67 (commented):
Hi @heinrich26 , pretty cool that you were able to make this work!

I take it you're basically emulating the mobile client to fetch the timed data? Since you're modifying the client info.

Could we replace the previous `get_lyrics` with this? I wouldn't want to have two methods for the same purpose in the API if one provides a superset of information.

Finally, to get this merged:

- please ensure the linters pass
- please add a test for the function

tests/mixins/test_browsing.py
```diff
@@ -164,9 +165,25 @@ def test_get_song_related_content(self, yt_oauth, sample_video):
...
+        assert isinstance(lyrics_song["lyrics"], list)
+        assert lyrics_song["hasTimestamps"] is True
+
+        # check the LyricLine object
+        song = lyrics_song["lyrics"][0]
+        assert isinstance(song, LyricLine)
+        assert isinstance(song.text, str)
+        assert isinstance(song.start_time, int)
+        assert isinstance(song.end_time, int)
+        assert isinstance(song.id, int)
```
sigma67, line 186: Not sure how useful these asserts are. Better to verify the length of the text or the time being greater than a specific value.

I.e., verify the usefulness of the actual values, not just the type

ytmusicapi/mixins/browsing.py
```diff
@@ -836,34 +882,180 @@ def get_song_related(self, browseId: str):
...
-            response, ["contents", *SECTION_LIST_ITEM, *DESCRIPTION_SHELF, *DESCRIPTION], True
-        )
-        lyrics["source"] = nav(
-            response, ["contents", *SECTION_LIST_ITEM, *DESCRIPTION_SHELF, "footer", *RUN_TEXT], True
-        )
 
-        return lyrics
+        if timestamps:
+            # restore the old context
+            self.context["context"]["client"] = copied_context_client  # type: ignore
```
sigma67, line 1031: type ignore should not be needed

ytmusicapi/mixins/browsing.py
```diff
@@ -836,34 +882,180 @@ def get_song_related(self, browseId: str):
...
+
+        lyrics: dict = {}
         if not browseId:
-            raise YTMusicUserError("Invalid browseId provided. This song might not have lyrics.")
+            raise YTMusicUserError(
+                "Invalid browseId provided. This song might not have lyrics.")
+
+        if timestamps:
+            # change the client to get lyrics with timestamps (mobile only)
+            copied_context_client = self.context["context"]["client"].copy()
```
sigma67, line 1021: This could be used in more places in the future, so I would suggest moving it to a separate utility function that returns a deepcopy of the context changed to the mobile context (maybe `get_mobile_context` next to initialize_context in helpers.py)

ytmusicapi/mixins/browsing.py
```diff
@@ -836,34 +882,180 @@ def get_song_related(self, browseId: str):
...
+        # we got lyrics with timestamps
+        if timestamps and (data := nav(response, TIMESTAMPED_LYRICS, True)) is not None:
+            assert isinstance(data, dict)
+
+            if not "timedLyricsData" in data:
+                return None
+
+            lyrics["lyrics"] = list(map(LyricLine.from_raw, data["timedLyricsData"]))
+            lyrics["source"] = data.get("sourceMessage")
+            lyrics["hasTimestamps"] = True
```
sigma67, line 1044: You could simplify this like so:

```
lyrics["hasTimestamps"] = timestamps and (data := nav(response, TIMESTAMPED_LYRICS, True))
if lyrics["hasTimestamps"]:
...
else:
...
```
Saves the explicit assignments.

ytmusicapi/mixins/browsing.py
```diff
@@ -836,34 +882,180 @@ def get_song_related(self, browseId: str):
...
+                    "hasTimestamps": True
+                }
 
         """
-        lyrics = {}
+
+    @overload
+    def get_lyrics(self, browseId: str, timestamps: Literal[True] = True) -> Optional[Lyrics|TimedLyrics]:
+        """
+        Returns lyrics of a song or video. When `timestamps` is set, lyrics are returned with
```
sigma67, line 932: The docstring is now duplicated in three places, please fix

Only the actual function should have the docstring, overloads don't need it

ytmusicapi/mixins/browsing.py
```diff
@@ -876,7 +1068,7 @@ def get_basejs_url(self):
         if match is None:
             raise YTMusicError("Could not identify the URL for base.js player.")
 
-        return YTM_DOMAIN + match.group(1)
+        return cast(str, YTM_DOMAIN + match.group(1))
```
sigma67, line 1071: please don't change unrelated functions

sigma67 (commented):
Thanks, looks much better! :)

As a general rule of thumb: Don't change more than necessary in a single PR. It makes it harder for the maintainer to review the changes (more time consuming), and harder to trace back potential future issues to specific PRs. It will also prolong the time it takes for the PR to get merged.

Note for this review: Please don't resolve threads, I will do so myself after verifying the changes

ytmusicapi/mixins/_protocol.py
```diff
@@ -17,15 +18,21 @@ class MixinProtocol(Protocol):
...
 
+    context: dict
+
     def _check_auth(self) -> None:
         """checks if self has authentication"""
+        ...
 
     def _send_request(self, endpoint: str, body: dict, additionalParams: str = "") -> dict:
         """for sending post requests to YouTube Music"""
+        ...
```
sigma67, line 29: The dots are not needed, please remove?

heinrich26: Pylance will report an error if not. 
(if you would remove the docstring, it would throw an error anyway, as the function is technically empty. See: https://docs.python.org/3/library/typing.html#typing.overload)

ytmusicapi/mixins/browsing.py
heinrich26: imo this is worse, because you assign to a dict, just to access it again. 🤷

ytmusicapi/mixins/browsing.py
heinrich26: this is technically not a type ignore, but it will supress the typechecker error: 
> copied_context_client might not be initialized

ytmusicapi/mixins/browsing.py
heinrich26: Agreed, however I wouldn't deepcopy the context, because someone might want to do weird stuff with references inside self.context. Using _shallow_-copy here, should structually restore everything to how it was before the call. (so I think)

a cool thing would be using a context manager for use `with`, like
```python
with yt.as_mobile():
    ...
# back to desktop
```
([contextlib](http://docs.python.org/library/contextlib.html)) makes a function a context-manager

ytmusicapi/mixins/browsing.py
```diff
@@ -271,13 +315,15 @@ def get_artist(self, channelId: str) -> dict:
             musicShelf = nav(results[0], MUSIC_SHELF)
             if "navigationEndpoint" in nav(musicShelf, TITLE):
                 artist["songs"]["browseId"] = nav(musicShelf, TITLE + NAVIGATION_BROWSE_ID)
-            artist["songs"]["results"] = parse_playlist_items(musicShelf["contents"])
+            artist["songs"]["results"] = parse_playlist_items(musicShelf["contents"]) # type: ignore
```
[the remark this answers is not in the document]
heinrich26: Pylance believes you're acessing a `None `value bc `browseId `was assigned to `None` before 🤷

ytmusicapi/mixins/browsing.py
```diff
@@ -24,6 +25,49 @@
...
+    def from_raw(cls, raw_lyric: dict):
+        """
+        Converts lyrics in the format from the api to a more reasonable format
+
+        :param raw_lyric: The raw lyric-data returned by the mobile api.
+        :return LyricLine: A `LyricLine`
+        """
+        text = raw_lyric["lyricLine"]
+        cue_range = raw_lyric["cueRange"]
+        start_time = int(cue_range["startTimeMilliseconds"])
```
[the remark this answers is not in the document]
heinrich26: sure, but it would also break if there was no cueRange or any other keys

ytmusicapi/mixins/_utils.py
```diff
@@ -1,9 +1,13 @@
 import re
 from datetime import date
+from typing import Literal
 
 from ytmusicapi.exceptions import YTMusicUserError
 
 
+OrderType = Literal['a_to_z', 'z_to_a', 'recently_added']
```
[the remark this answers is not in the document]
heinrich26: yes

ytmusicapi/__init__.py
```diff
@@ -2,14 +2,16 @@
 
 from ytmusicapi.setup import setup, setup_oauth
 from ytmusicapi.ytmusic import YTMusic
+from .mixins.browsing import Lyrics, TimedLyrics, LyricLine
```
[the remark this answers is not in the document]
heinrich26: I was thinking similar, but then you'll need to create a submodule for these classes to be accessible

tests/mixins/test_browsing.py
heinrich26: The text could easily be empty, if there was a pause. One character lines are quite common at the beginning of a song (there's that `note character`)

ytmusicapi/mixins/_protocol.py
sigma67: This project uses mypy, so please don't worry about pylance findings. For mypy it's ok to have a docstring in place of `...`

It's a matter of personal preference - for me the docstring has a purpose and fulfills the requirement of no empty function body, so I like it better than the nondescript `...`

sigma67: Side note: I have triggered the lint jobs so you can see the mypy and ruff findings remaining. 

I recommend taking a look at https://github.com/sigma67/ytmusicapi/blob/main/CONTRIBUTING.rst

where pre-commit is described - if that passes, the lint jobs should pass as well.

ytmusicapi/mixins/browsing.py
sigma67: ok, could you either revert the change or add the return type then? I'd prefer the latter

ytmusicapi/mixins/browsing.py
sigma67: ok, fine with me

ytmusicapi/mixins/browsing.py
sigma67: Interesting idea, I like it. But it won't be pretty to wrap the entire function body in a with statement?

And wouldn't it cause problems during parallel execution? If I call a two endpoint with the same `YTMusic` instance at the same time, wouldn't one of them break if you perform a shallow copy?

ytmusicapi/__init__.py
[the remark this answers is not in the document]
sigma67: You can already access them: ``from ytmusicapi.mixins.browsing import TimedLyrics``

It's similar for https://github.com/sigma67/ytmusicapi/blob/main/ytmusicapi/parsers/podcasts.py

Although with #621 in mind, it might be wise to move them to a `ytmusicapi.models` module (one for `podcasts` and one for `lyrics`, I guess)

ytmusicapi/mixins/browsing.py
heinrich26: oh yes, I see what you're meaning. I kinda misunderstood your deepcopy, but if you deepcopied the context only for the operation it's totally fine and thread safe. Will do that.

ytmusicapi/mixins/uploads.py
```diff
@@ -19,16 +19,16 @@
...
 
 class UploadsMixin(MixinProtocol):
-    def get_library_upload_songs(self, limit: Optional[int] = 25, order: Optional[str] = None) -> list[dict]:
+    def get_library_upload_songs(self, limit: Optional[int] = 25, order: Optional[OrderType] = None) -> list[dict]:
         """
         Returns a list of uploaded songs
 
         :param limit: How many songs to return. `None` retrieves them all. Default: 25
-        :param order: Order of songs to return. Allowed values: 'a_to_z', 'z_to_a', 'recently_added'. Default: Default order.
+        :param order: Order of songs to return. Allowed values: `a_to_z`, `z_to_a`, `recently_added`. Default: Default order.
```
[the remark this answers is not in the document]
heinrich26: moved to another PR

heinrich26:
All issues should have been resolved

heinrich26:
finally used pre-commit to fix those formatting complaints 🤷‍♂️

sigma67:
Hi @heinrich26 , I am moving this over to #693 as merging is somehow blocked here

PR #664, opened 2024-10-15
replace Requests http client by Niquests

Ousret:
This PR effectively replace the http client Requests for [Niquests](https://github.com/jawah/niquests).
Niquests is a drop-in replacement for Requests that is no longer under feature freeze.

This new client support HTTP/2, and HTTP/3 by default and offers both sync and async interfaces. It supports all the latest shiny features you would expect from an http client.

If you were interested on merging, I will be interested to propose a followup PR that will propose a script that generate
the async part of ytmusicapi automatically. If you want to.

*disclaimer:* _I maintain Niquests._


sigma67:
Sorry, but from my perspective there are no benefits.There are no features missing in requests for the purposes of this library. 

In addition a lot of risk would be introduced by switching to something less widely used than requests which has been stable for years and is PSF supported.

sigma67:
Also note that this broke half of all tests, seems it's not just a drop-in

https://github.com/sigma67/ytmusicapi/pull/664/checks

Ousret:
You came back to take a deeper lock.
This is somewhat interesting.

> There are no features missing in requests for the purposes of this library.

That I may understand. But it is not entirely true. I would not propose this if your community wouldn't benefit from it.
Browsing your issue tracker shows horrible things like `self.code = await self.hass.async_add_executor_job(self.oauth.get_code)`
Native async would be a nice addition. According to me.

Then, you should know that YT admins may, without notice, kill HTTP/1 requests to make this lib useless. So, HTTP/2, and HTTP/3 would be nice to gain more confidence on that topic.

> In addition a lot of risk would be introduced by switching to something less widely used than requests which has been stable for years

That I've heard many times. This thinking is tautological. Let's summarize what you said:

Old + Big usage = Better ?

So, if given a choice (for free) between a "Toyota Corolla from 2003" and a "Tesla Model X from 2023" you would pick the first by that logic. You would also choose Google Chrome over others without question. (and so on..)

Ok, so there I am not saying that Niquests is bug-free, but it's in a pretty good place right now.
We should be concerned more by the ability of someone to respond to bug than on his ability to have none from the start. And today, if you pick a look at Requests, it is bad.
Feel free to inspect what we have done and how we treat issues (both Requests and us)

Progress can't happen otherwise.

> Also note that this broke half of all tests, seems it's not just a drop-in

You closed the PR in minutes, so I didn't get the chance of looking further.
The errors aren't bad, and they are now fixed the branch.

If you're curious, re-open and re-rerun them. hint: `CaseInsensibleDict` isn't made for setting storage, but only http headers. It was a mistake from Requests era to allow garbage as values.

regards,

sigma67:
Thank you for the eloquent and detailed response, appreciate it. To address your points:

- I will admit that async has been a somewhat frequently requested feature for this library. But I don't yet see how niquests would solve that issue, and autogenerating code to solve it seems like a strange approach to me (albeit I have not looked into this further)
- How much merit does the " YT admins may, without notice, kill HTTP/1 requests to make this lib useless" argument really have? Is there any indication that this may actually happen?
- "Old + Big usage = Better ?" No, this is not my argument. Better is hard to define anyway. What matters for me only, and I will repeat that, is stability. Requests has done the job since this library was created almost 4 years ago with 0 breaking changes. In contrast, the first version of niquests was released a year ago (and you admit it is not bug free - fair). 
- If requests ceases to do the job,I will swap it for a decent replacement. Keep in mind that niquests is not the only alternative and the choice must be based on what is the best fit for this library (and ideally least maintenance effort - as a fork of requests you have one up here)

I will keep your project in mind, but at this point it is too early for me to make a decision that is not prompted by a need

Ousret:
> autogenerating code to solve it seems like a strange approach to me

You are not alone on this take. But unfortunately, no other way emerge to solve this.
There is no reliable way to have a single code base and do both sync and async.

Basically here, you'll need to duplicate code and all the way down from the http calls up to the user front call, you'll need to convert async/await(able).

I've already contributed to that goal at https://github.com/grafana-toolbox/grafana-client/blob/main/script/generate_async.py
The code between async and sync doesn't change, so to relieve you of the burden of updating the async part frequently or ask contributor to do so every time, you may prefer this option.

> Is there any indication that this may actually happen?

There is no official statement (figure...) but all over internet, the number of reports about bans start to grow. You have yourself some complaint about it.

Not only Google, but major provider like Cloudflare use the way you connect to them (TLS layer, http headers, protocol used, cipher, etc...) to determine how likely you are NOT a traditional browser. (see https://niquests.readthedocs.io/en/latest/user/advanced.html#tls-fingerprint-like-ja3 for more info)

This is how they target and act upon library/client like yours.

I am extremely confident that they will kill HTTP/1 soon. Some services (major) in my home country started to progressively kill HTTP/1, this was the starting point of the development for Niquests.

> Requests has done the job since this library was created almost 4 years ago with 0 breaking changes.

To you, yes, no breaking. But for other major libraries out there, there are major breakage happening while they are in maintenance mode only. And there's still major fixes awaited by the community, sometimes for months or years.

> [this might be one of those](https://github.com/sigma67/ytmusicapi/actions/runs/11371348924/job/31633212397?pr=664#step:7:338)).

You are right, this happen to be a minor issue. The Google QUIC implementation seems to be aggressive toward keepalive/ACKs.
We've applied a fix for that.

> If requests ceases to do the job

Don't wait for complete "out-of-order" situation. Maybe try to be proactive about this one. Keep in mind that Requests stated that the project is already half-dead.

>  but at this point it is too early for me to make a decision that is not prompted by a need

I understand completely. Take your time.
If you need further collaboration, I'll be there.

Until then,

Regards,

sigma67:
> You are right, this happen to be a minor issue. The Google QUIC implementation seems to be aggressive toward keepalive/ACKs.
We've applied a fix for that.

Could you point out the mentioned fix?

> Keep in mind that Requests stated that the project is already half-dead.

Could you link the source for this please?


Additionally, since your project claims performance benefits (up to 3x faster than requests), I had a look at the test runtime. It seems worse than with requests:

- reference runtime from merged PR 3m55s: https://github.com/sigma67/ytmusicapi/actions/runs/11325189047/job/31491559360
- niquests runtime 5m 20s: https://github.com/sigma67/ytmusicapi/actions/runs/11388349952/job/31685095011

Granted that most of the runtime is to blame on server-side processing. Nevertheless it still seems like a significant performance regression at first glance. Any clue?

Ousret:
> Could you point out the mentioned fix?

Of course. See https://github.com/jawah/urllib3.future/pull/161 and https://github.com/jawah/urllib3.future/pull/162

> Could you link the source for this please?

- https://blog.ian.stapletoncordas.co/2024/02/a-retrospective-on-requests
- https://vorpus.org/blog/why-im-not-collaborating-with-kenneth-reitz/
- https://x.com/kennethreitz42/status/1723329022422110444
- https://github.com/psf/requests/pull/
  - See n°6731 for example
  - Or n°6568
  - Or https://github.com/psf/requests/issues/6665
  - ....
- https://medium.com/@ahmed.tahri/revived-the-promise-made-six-years-ago-for-requests-3-37b440e6a064

There is more, but, you will excuse me to not collect the hundreds of demonstration for it.

> It seems worse than with requests:

Short story: no multiplexing used, only pure synchronous. re-do QUIC handshake hundred of times using Post-Quantum Security, don't reuse session, etc...

See https://niquests.readthedocs.io/en/latest/community/faq.html#why-http-2-and-http-3-seems-slower-than-http-1-1

To witness the speed up in CI, that's tricky, because you'll need to instantiate once the niquests.Session() and leverage pytest concurrency, then you'll see improvement for sure, up to 2 times faster. (x3 times is reserved to async or manual multiplexing aka. lazy responses)

To be clear, our claim about 3x faster is accompanied with a reproducer / specific scenario. If you use HTTP/2+ as you use HTTP/1, it will be slower because you don't leverage his potential.

On the latest run, we get:
- ======= 1 failed, 104 passed, 1 skipped, 1 retried in 218.81s (0:03:38) ========

so, not so significant after all.

The failure here seems unrelated to us: 
- pr664 https://github.com/sigma67/ytmusicapi/actions/runs/11435641864/job/31811422660?pr=664
- pr662 https://github.com/sigma67/ytmusicapi/actions/runs/11426051835/job/31788490025?pr=662

regards,

Ousret:
It seems that we are facing more ban from YT. The phenomena is getting more and more important.
(Looking at google recent results + GH issues mentioning YT+ban)

maybe it worth to pursue that PR and getting HTTP/3 mainstream.

Ousret:
Going to close this one for now.
If you want to pursue this in the future, feel free to ping me any time.

Regards,

PR #665, opened 2024-10-16
Docstring cleanup

heinrich26:
Replaced occurences of \`\` / ' with \` in Docstrings


sigma67 (changes requested):
There is a semantic difference between ` and ``

https://docutils.sourceforge.io/docs/user/rst/quickref.html#inline-markup

In this case, double backtick is correct. Using single backtick while show up differently in the rendered HTML markup

```
`Test` vs ``test``.
```

image

sigma67:
@heinrich26 still interested in merging this?

PR #667, opened 2024-10-28
get_playlist: join multiple title runs

sgvictorino:
* Return complete titles when split on time-like strings:

```diff
>>> yt.get_playlist("PLaZPMsuQNCsWn0iVMtGbaUXO6z-EdZaZm")["title"]
-03 Jan
+03 Jan 12:09
```

* `test_get_playlist`: assert more props have the expected value/shape


sigma67:
@sgvictorino you may have forgotten to `git add` the files :)

Issue #679, opened 2024-11-16
oauth fixed in 1.9.0 - requires client_id / client_secret


sigma67:
As discussed in #676 , oauth-based authentication has been disabled for the web client by Google. It is unknown if it will work again in the future.

Therefore, it is advised to use browser-based authentication instead: https://ytmusicapi.readthedocs.io/en/stable/setup/browser.html

Contributions towards improving the browser-based setup option are welcome as always. Please open an issue to discuss the proposed change beforehand.

sigma67:
For discussion, please see here: https://github.com/sigma67/ytmusicapi/discussions/682

sigma67:
It looks like there will be a way to get it working again. Watch out for an update in the coming weeks. 

However, unlike previously it will require you to get a client secret and id from the YouTube Data API.

PR #688, opened 2024-12-13
Add optional parameters for OAuth client ID and secret to setup

fischcode:
Use your own Google API client ID and client secret in setup by calling `ytmusicapi oauth --client_id <your Google API client ID> --client_secret <your Google API client secret>` to avoid HTTP 400: Bad Request errors (issue #679 )
See [developers.google.com/youtube/registering_an_application](https://developers.google.com/youtube/registering_an_application) for instructions on how to create your own youtube OAuth credentials.


sigma67 (commented):
Thank you for for figuring this out and submitting the PR. However, there are still some more improvements needed:

- The argparse args you added apply to both setup types, although they don't matter for the `browser` option. This will lead to a misleading CLI (i.e. try running ytmusicapi browser --help). They should only apply to the oauth option. Since the interface is interactive anyway you could also circumvent this by interactively asking for TV auth (deprecated) or client id/secret (recommended) - then request all the info via `input()`. Both solutions are valid imo.
- Deprecate the current TV oauth docs and include info on how it should be done currently. This would mean adding guidance on how to correctly create the ID/Secret via Google Cloud. Could be done in a separate PR

sigma67:
Which type of OAuth 2.0 client/id secret are you creating? "TV and limited input" or something else? I'm still getting 401 unauthorized here

fischcode:
Hi
I added subparsers for oauth / browser and validation if oauth is chosen as requested. Hope this meets your expectations.
However, I am reluctant to write instructions on how to create API keys at this time, as I haven't yet figured out the minimum scope required for all ytmusicapi features to work and Google appears to be in the process of changing the UI for key creation.
Here is one page with instructions ("API Key Creation") that you might want to test and copy into the ytmusicapi documentation if successful: https://github.com/Kuschranada/plugin.video.youtube/wiki/Personal-API-Keys

PR #689, opened 2024-12-14
Move songlist parsing get_history

luisignaciocc:
Fix #687 and #686

PR #691, opened 2024-12-16
Fix get_history

sigma67:
actual fix for #689

### DIFF
--- a/tests/auth/test_oauth.py
+++ b/tests/auth/test_oauth.py
@@ -63,6 +63,7 @@ class TestOAuth:
         oauth_file.close()
         Path(oauth_file.name).unlink()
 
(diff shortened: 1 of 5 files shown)


sigma67:
fyi @luisignaciocc your PR only circumvented the issue but did not fix it entirely.

The root cause is this new pin menu entry added by Google to history items.

This call was relying on the last entry in the list being the one with the `feedbackToken` to remove something from the history.

If you remove the menu parsing entirely this will not get parsed, breaking things like `remove_history_items` because there is no longer a way to get the required `fefedbackToken` for it.

luisignaciocc:
Oops, my bad! I didn’t have the full context about everything that was involved in that change.

PR #696, opened 2024-12-17
get_playlist: support audio playlists

sgvictorino:


sigma67:
You're trying to support albums with `get_playlist`.

Albums are meant to be retrieved with `get_album`. 

I don't see the value in supporting the same content with two different endpoints.

sigma67:
I realize that right now it's maybe not convenient to call `get_album_browse_id(audioPlaylistId)` before calling `get_album(browseId)`, but that's just how the API works

sigma67:
Please also following CONTRIBUTING.rst for more involved changes so wasted effort can be prevented:

https://github.com/sigma67/ytmusicapi/blob/main/CONTRIBUTING.rst#pull-requests

```
Please open an issue before submitting, unless it's just a typo or some other small error.
```

sigma67:
Hmm I guess that's more of a direct consequence of grabbing an album. Everything inside an album should always be "ATV":

https://ytmusicapi.readthedocs.io/en/stable/faq.html#which-videotypes-exist-and-what-do-they-mean

"High quality song uploaded by original artist with cover image"

Unless you have another argument that would require this PR I would go ahead and close this, since everything can be achieved using existing means.

sigma67:
Ok fair, I guess they do actually return different content. I would accept this if you could move the separate parsing for audio playlists into a separate parsing function. 

Don't make the audioPlaylist codepath dependent on a property of the data, instead make it dependent on the playlistId. If it starts with OLA, use the separate function for parsing the data.

This should avoid mangling the code with too many if else statements while keeping it somewhat readable.

ytmusicapi/mixins/playlists.py
```diff
@@ -190,6 +192,42 @@ def get_playlist(
...
+
+        section_list = nav(response, [*TWO_COLUMN_RENDERER, "secondaryContents", *SECTION])
+        content_data = nav(section_list, [*CONTENT, "musicPlaylistShelfRenderer"])
+
+        playlist["id"] = nav(
+            content_data, [*CONTENT, MRLIR, *PLAY_BUTTON, "playNavigationEndpoint", *WATCH_PLAYLIST_ID]
+        )
+        playlist["trackCount"] = nav(content_data, ["collapsedItemCount"])
+
+        playlist["tracks"] = []
```
sigma67, line 214: can be moved to the declaration in l.196

ytmusicapi/mixins/playlists.py
```diff
@@ -190,6 +192,42 @@ def get_playlist(
...
+        if "contents" in content_data:
+            playlist["tracks"] = parse_playlist_items(content_data["contents"])
+
+            parse_func = lambda contents: parse_playlist_items(contents)
+            if "continuations" in content_data:
+                playlist["tracks"].extend(
```
sigma67, line 220: this is not covered.

ytmusicapi/mixins/playlists.py
```diff
@@ -190,6 +192,42 @@ def get_playlist(
         playlist["duration_seconds"] = sum_total_duration(playlist)
         return playlist
 
+    def _parse_audio_playlist(self, response: dict, limit: Optional[int], request_func) -> dict[str, Any]:
```
sigma67, line 195: Please move to `parsers/playlists.py`. Sorry if I wasn't clear here. 

The idea is that the mixins contain only the public API and everything related to parsing goes into a parsers/xx file.

sigma67 (commented):
Looks good, almost finished

PR #698, opened 2024-12-19
search: exclude "Single"/"EP" text from album artist list

sgvictorino:
```diff
 next(r for r in yt.search("Eminem Houdini", filter="albums") if r["title"] == "Houdini")["artists"]

 [
-    {'name': 'Single', 'id': None},
     {'name': 'Eminem', 'id': 'UCedvOgsKFzcK3hA5taf3KoQ'}
 ]
```


sigma67:
This is fine but the translations are incomplete. 

It's fairly easy to retrieve the correct values automatically by looping over the `YTMusic` instance with the different language values and printing the single/ep value. Then add it to the po files. Keep in mind that they need to be compiled, see `locales/README`.

Please also add tests.

sigma67 (approved):
Thank you for your contribution!

PR #700, opened 2024-12-22
fix(setup): Use resolve() before invoking as_uri()

### LINKED ISSUE #699: 1.9.0: ytmusicapi oauth results in "ValueError: relative path can't be expressed as a file URI"
**Describe the bug**

```
$ ytmusicapi oauth
  File "venv/bin/ytmusicapi", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "venv/lib/python3.11/site-packages/ytmusicapi/setup.py", line 78, in main
    print(f"Creating {Path(filename).as_uri()} with your authentication credentials...")
                      ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/pathlib.py", line 567, in as_uri
    raise ValueError("relative path can't be expressed as a file URI")
ValueError: relative path can't be expressed as a file URI
```

**To Reproduce**

run the code above in python3.11 or a small reproducer:

```
>>> from pathlib import Path
>>> Path('oauth.json').as_uri()
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.11/pathlib.py", line 567, in as_uri
    raise ValueError("relative path can't be expressed as a file URI")
ValueError: relative path can't be expressed as a file URI
```

which is what's done in https://github.com/sigma67/ytmusicapi/blob/main/ytmusicapi/setup.py#L77-L78

apastel:
If no `--file` argument is provided to setup, `as_uri()` throws a ValueError when it tries to get the relative path to print to the console. Using `resolve()` will work in both case (`--file` provided or not provided).

Fixes #699

### DIFF
--- a/ytmusicapi/setup.py
+++ b/ytmusicapi/setup.py
@@ -75,7 +75,7 @@ def parse_args(args):
 def main() -> Union[RefreshingToken, str]:
     args = parse_args(sys.argv[1:])
     filename = args.file.as_posix() if args.file else f"{args.setup_type}.json"
-    print(f"Creating {Path(filename).as_uri()} with your authentication credentials...")
+    print(f"Creating {Path(filename).resolve().as_uri()} with your authentication credentials...")
     if args.setup_type == "oauth":
         if args.client_id is None:
             args.client_id = input("Enter your Google Youtube Data API client ID: ")


sigma67 (commented):
Not instead, in addition. Don't change the behavior of the print statement - with this change the link won't be clickable in terminal

Please also add a test

apastel:
I'm inclined to say the test is that you run `ytmusicapi browser|oauth` with a relative file path and that it doesn't throw an error anymore. I spent some time trying to test the argparse CLI and it's not very intuitive like it is with `click`, the CLI library that I'm used to, and there aren't yet any examples in the project of tests that actually execute the `main()` function without mocking it. I don't have the bandwidth to dive into it today beyond what I tried already.

sandrotosi:
in my local installation i ended up just removing `as_uri()`, then i found out about this PR lol anyhow i think the cli to setup authentication being broken is more important than a file link being clickable and should be fixed fast, just my 2c

PR #724, opened 2025-01-16
Check if auth string is JSON before trying to load from file.

0kommanix:


ytmusicapi/ytmusic.py
```diff
@@ -123,7 +123,7 @@ def __init__(
         if self.auth is not None:
             self.oauth_credentials = oauth_credentials
             auth_path: Optional[Path] = None
-            if isinstance(self.auth, str):
+            if not (auth_str.startswith("{") and (auth_path := Path(auth_str)).is_file():
```
sigma67, line 126:
```suggestion
            if isinstance(self.auth, str) and not (auth_str.startswith("{"):
```

Issue #733, opened 2025-01-31
docs build fails in 1.10


antonio-rojas:
**Describe the bug**
Docs build fails in 1.10 with
```
make: Entering directory '/build/python-ytmusicapi/src/ytmusicapi/docs'
Makefile:20: *** missing separator.  Stop.
make: Leaving directory '/build/python-ytmusicapi/src/ytmusicapi/docs'
```

**To Reproduce**
1. Run `make -C docs`

**Additional context**
make 4.4.1, sphinx 8.1.3

PR #737, opened 2025-02-03
Fix docs build

antonio-rojas:
Make needs tabs, not spaces

Fixes https://github.com/sigma67/ytmusicapi/issues/733


sigma67:
I prefer the Python way 🥹

https://peps.python.org/pep-0008/#tabs-or-spaces

PR #738, opened 2025-02-03
cache visitor ID on first use

sgvictorino:
```diff
import ytmusicapi
import timeit

yt = ytmusicapi.YTMusic()
for _ in range(2):
  print(round(timeit.timeit(lambda: yt.get_album('MPREb_4pL8gzRtw1p'), number=1) * 1000))

-710
-593
+753
+146
```

### DIFF
--- a/ytmusicapi/ytmusic.py
+++ b/ytmusicapi/ytmusic.py
@@ -4,7 +4,7 @@ import locale
 import time
 from collections.abc import Iterator
 from contextlib import contextmanager, suppress
-from functools import partial
+from functools import cached_property, partial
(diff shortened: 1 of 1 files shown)


sigma67:
Wouldn't it make more sense to move it into `base_headers` if it shouldn't be updated?

sigma67:
It is currently not used for that. My suggestion is to move the fetching of the visitor ID into base_headers and use ``@cached_property`` on ``base_headers``.

sigma67:
Does the suggested solution work for you? Sorry for the breakage caused by the refactor.

PR #755, opened 2025-03-29
Fix singles being renamed to "Singles & EPs"

### LINKED ISSUE #754: Singles missing from artist listing
**Describe the bug**
When getting info about an artist using `get_artist`, there are no singles, even if that artist does have singles

**To Reproduce**
```python
ytmusic: YTMusic = YTMusic()
artist = ytmusic.get_artist('UC-CcyIM_seGnGL5-2Fsppow')
print(artist['singles']) # KeyError
```

**Additional context**
I am using the latest version on main.

fabi321:
This Fixes a bug, that singles where renamed to `Singles & EPs`, and thus not found any more.

Fixes #754


fabi321:
I didn't touch the translation files so far, not sure what is needed there. Alternative fix would be to just look if `singles` is contained in the category by changing
https://github.com/sigma67/ytmusicapi/blob/55b362e4dc6722800b54c5adaa5ad69019e72897/ytmusicapi/parsers/i18n.py#L70
to
```python
                and category_local.lower() in nav(r, CAROUSEL + CAROUSEL_TITLE)["text"].lower()
```

fabi321:
After looking through the translations by @sgvictorino, the quick'n'dirty version seems to work for most languages, but not some.

ytmusicapi/locales/es/LC_MESSAGES/base.po
```diff
@@ -92,3 +92,6 @@ msgstr "últimos episodios"
 #: parsers/i18n.py:62
 msgid "podcasts"
 msgstr "pódcasts"
+
+#~ msgid "singles"
```
sigma67, line 96: Please remove this

sigma67 (commented):
Hi @fabi321 , thanks for the contribution.

Looks good to me except for the failing test. Why do you think it's not passing for some languages?

fabi321:
The failed test seemed to have been a one off.

PR #773, opened 2025-05-25
Changed the path where the token is saved (to use the API in Vercel)

moguism:
I was trying to use the API (thank you for making it btw!) in a project deployed with Vercel. It was used to show an image with the last song you listened to, but almost everytime I wanted to see the image, I got an Internal Server Error.
It turns out that Vercel filesystem is read only, so everytime the API tries to save the token, it gives an error. The specific problem was found in the changed function, and I just needed to store the token in a temp file to solve it.
I looked through all the code and it seems that the "store_token" function isn't used anywhere else, so this change shouldn't cause any issue.

### DIFF
--- a/ytmusicapi/__init__.py
+++ b/ytmusicapi/__init__.py
@@ -1,17 +1,17 @@
 from importlib.metadata import PackageNotFoundError, version
 
-from ytmusicapi2.auth.oauth.credentials import OAuthCredentials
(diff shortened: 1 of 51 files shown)


sigma67:
We're certainly not going to place it in a fixed unix-only location.

You can already determine the output path by passing in the file via `auth` param: https://ytmusicapi.readthedocs.io/en/stable/reference/ytmusic.html

PR #791, opened 2025-07-28
get_album: distinguish between "other versions" and "releases for you"

sgvictorino:
- adds `related_recommendations`, a list of albums in "Releases for you"
- `other_versions` no longer contains those recommendations when its carousel is positioned after "Releases for you"
- `audioPlaylistId` is no longer `None` when `parse_album` is called with a Premium account (`test_get_album_other_versions` can now be used to assert this)


ytmusicapi/parsers/albums.py
```diff
@@ -85,3 +85,8 @@ def parse_album_header_2024(response: JsonDict) -> JsonDict:
         album["likeStatus"] = parse_like_status(service)
 
     return album
+
+
+def parse_album_playlistid_if_exists(data: JsonDict | None) -> str | None:
+    """the content of the data changes based on whether the user is authenticated or not"""
+    return nav(data, WATCH_PID, True) or nav(data, WATCH_PLAYLIST_ID, True) if data else None
```
sigma67, line 92: This difference is caused by an A/B rollout? It's terribly confusing to me when they use `WATCH_PID` vs `WATCH_PLAYLIST_ID`

PR #792, opened 2025-07-28
search: fix episode parsing

sgvictorino:
- `parse_top_result`: handle episode data
- `parse_search_result`: ensure `resultType` is "episode", not "video"
- `parse_search_result`: offset runs to skip the "Episode" specifier


ytmusicapi/parsers/search.py
```diff
@@ -191,12 +201,13 @@ def parse_search_result(
 
     if result_type in ["episode"]:
         flex_item = get_flex_column_item(data, 1)
-        has_date = int(len(nav(flex_item, TEXT_RUNS)) > 1)
+        runs = nav(flex_item, TEXT_RUNS)[default_offset:]
```
sigma67, line 204: The new/fix part is the added default offset here?

PR #798, opened 2025-08-03
get_home: fix song artist parsing

sgvictorino:
extra text is now excluded from `artists`:

- `get_home`/`get_song_related`: play counts in certain track carousels
- `get_home`: "Song" specifiers for audio tracks

`get_explore` song charts are also no longer missing `album` data

### DIFF
--- a/ytmusicapi/parsers/_utils.py
+++ b/ytmusicapi/parsers/_utils.py
@@ -7,6 +7,8 @@ from typing import ParamSpec, TypeVar
 from ytmusicapi.navigation import *
 from ytmusicapi.type_alias import JsonDict, JsonList
 
+from .constants import DOT_SEPARATOR_RUN
+
 P = ParamSpec("P")
 R = TypeVar("R")
 
@@ -69,7 +71,7 @@ def get_fixed_column_item(item: JsonDict, index: int) -> JsonDict | None:
 
 def get_dot_separator_index(runs: JsonList) -> int:
     try:
-        index = runs.index({"text": " • "})
+        index = runs.index(DOT_SEPARATOR_RUN)
     except ValueError:
         index = len(runs)
 
--- a/ytmusicapi/parsers/songs.py
+++ b/ytmusicapi/parsers/songs.py
@@ -3,6 +3,7 @@ import re
 from ytmusicapi.type_alias import JsonDict, JsonList
 
 from ._utils import *
+from .constants import DOT_SEPARATOR_RUN
 
 
 def parse_song_artists(data: JsonDict, index: int) -> JsonList:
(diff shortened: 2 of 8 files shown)


ytmusicapi/parsers/explore.py
```diff
@@ -43,18 +43,6 @@ def parse_chart_artist(data: JsonDict) -> JsonDict:
     return parsed
 
 
-def parse_trending_song(data: JsonDict) -> JsonDict:
```
sigma67, line 46: Why is this removed?

ytmusicapi/parsers/songs.py
```diff
@@ -21,7 +21,13 @@ def parse_song_artists_runs(runs: JsonList) -> JsonList:
     return artists
 
 
-def parse_song_runs(runs: JsonList) -> JsonDict:
+def parse_song_runs(runs: JsonList, api_result_types: list[str] = []) -> JsonDict:
```
sigma67, line 24: This is not so nice because we have to pass this information (api_result_types) through a method chain involving several functions. Can we do better or is there an alternative approach?

sigma67 (commented):
Overall question: Is all of this related to `get_home` ? Since your also touching `parse_chart_song`

ytmusicapi/parsers/songs.py
```diff
@@ -21,9 +21,18 @@ def parse_song_artists_runs(runs: JsonList) -> JsonList:
...
-def parse_song_runs(runs: JsonList) -> JsonDict:
+def parse_song_runs(runs: JsonList, skip_type_spec: bool = False) -> JsonDict:
+    """
+    :param skip_type_spec: ignore data preceding artist text runs, which may be a type specifier like "Song", "Single", or "Album"
+    """
+
     parsed: JsonDict = {"artists": []}
-    for i, run in enumerate(runs):
+
+    for i, run in reversed(list(enumerate(runs))):
```
sigma67, line 31: I don't think we should reverse the list in general. 

I suspect it is done because you want to check the last item and abort. I would suggest putting that check before the loop.

ytmusicapi/parsers/songs.py
```diff
@@ -21,9 +21,18 @@ def parse_song_artists_runs(runs: JsonList) -> JsonList:
     return artists
 
 
-def parse_song_runs(runs: JsonList) -> JsonDict:
+def parse_song_runs(runs: JsonList, skip_type_spec: bool = False) -> JsonDict:
+    """
+    :param skip_type_spec: ignore data preceding artist text runs, which may be a type specifier like "Song", "Single", or "Album"
```
sigma67, line 26: Please add some more detail to the message - i.e. what the specific use case for this parameter is, why it exists

ytmusicapi/parsers/_utils.py
```diff
@@ -67,9 +67,12 @@ def get_fixed_column_item(item: JsonDict, index: int) -> JsonDict | None:
     return typing.cast(JsonDict, item["fixedColumns"][index]["musicResponsiveListItemFixedColumnRenderer"])
 
 
+DOT_SEPARATOR_RUN = {"text": " • "}
```
sigma67, line 70: I think it's a bit unexpected to place this in the middle of `_utils`.

If you want to keep it please create a separate file `parsers/constants.py`.

sigma67 (commented):
Nice, looks pretty clean now to me. I think the changeset looks bigger than needed because you have some formatting leftovers due to repeated changes.

Just some minor nitpicks, then we're good to go

ytmusicapi/parsers/songs.py
sigma67: That looks good to me, go ahead

PR #803, opened 2025-08-21
Added translation for Czech language

DJSweder:


sigma67:
Thank you for the contribution @DJSweder !

PR #819, opened 2025-09-29
distinguish "Save to library" and "Pin to Listen again" feedbackTokens

### LINKED ISSUE #771: Issues removing song from library using edit_song_library_status
I'm using `edit_song_library_status` to remove certain songs from the library. 

Steps to reproduce:
1. Use `get_library_songs(limit=1000, validate_responses=True)` to get songs in the library. 
2. Extract `track['feedbackTokens']['remove']` for tracks you want to remove
3. Use` edit_song_library_status(feedback_tokens)` to remove songs from the library
4. Use `get_library_songs()` again => The removed songs will also be returned

Thanks for looking into this. Let me know if I'm using the API incorrectly

sgvictorino:
Closes #771

- adds `pinnedToListenAgain` and `listenAgainFeedbackTokens`

- also refactors `get_history` to match on `iconType`
(`REMOVE_FROM_HISTORY`) instead of returning the first feedback token

### DIFF
--- a/tests/mixins/test_search.py
+++ b/tests/mixins/test_search.py
@@ -136,7 +136,7 @@ class TestSearch:
         assert len(results[0]["author"]) > 0
 
     def test_search_top_result_episode(self, yt_auth):
-        results = yt_auth.search("124. making meetings meaningful pt. 1 stanford")
+        results = yt_auth.search('"124. Making Meetings Meaningful, Pt. 1" Stanford')
         assert results[0]["category"] == "Top result"
         assert results[0]["resultType"] == "episode"
         assert results[0]["videoId"] == "KNkyHCLOr1o"


sigma67:
Thank you! I will have a look as soon as I am able. Will also need to replace some of the oauth tests with browser ones for now.

sigma67:
``test_get_album`` is still failing unfortunately

ytmusicapi/parsers/songs.py
```diff
@@ -94,24 +95,72 @@ def parse_song_album(data: JsonDict, index: int) -> JsonDict | None:
...
+
+    Example::
+
+        {
+            "inLibrary": true,
+            "feedbackTokens": {
+                "add": "...",
+                "remove": "..."
+            },
+            "pinnedToListenAgain": true,
```
sigma67: This one doesn't show up in the function docstrings in the mixins.

We should also add documentation on how these tokens are expected to be used

I.e. a FAQ entry on "What is the purpose of ``feedbackTokens`` vs ``listenAgainFeedbackTokens``?"

tests/mixins/test_browsing.py
sigma67: This assertion is still failing but I don't know why. Locally it passes for me.

sgvictorino: I think they were A/B testing the "Save to library" button icon.

sigma67 (approved):
The failing test is clearly unrelated so let's go ahead with this one

PR #848, opened 2025-12-22
feat(album): add ability to fetch 'clean' songs instead of 'dirty' songlist

kaneryu:
I'm actually not sure what to call this....

The basic issue is that when you call get_album the returned song list sometimes substitutes songs with music videos, lyric videos, etc etc, with different titles when you call get_song on the songid. This is a huge, very annoying issue directly in YouTube music too, and can make it impossible to access some extended versions of songs (they keep pointing towards the non extended music videos). This adds the ability to *always* fetch the 'clean' songs, that is, the youtube video's title is just song title, the video itself is just the song cover, etc. If you know what this is actually called, please enlighten me😅

### DIFF
--- a/tests/mixins/test_browsing.py
+++ b/tests/mixins/test_browsing.py
@@ -188,6 +188,11 @@ class TestBrowsing:
         assert variant["artists"][2] == {"id": "UCb7jnkQW94hzOoWkG14zs4w", "name": "D-Block Europe"}
         assert variant["audioPlaylistId"] is not None
 
+    def test_get_album_audio_tracks_only(self, yt):
+        album = yt.get_album("MPREb_jKN24KF3kFb", audio_tracks_only=True)
+        regular_album = yt.get_album("MPREb_jKN24KF3kFb", audio_tracks_only=False)
+        assert album["tracks"] != regular_album["tracks"]
+
     def test_get_song(self, config, yt, yt_oauth, sample_video):
         song = yt_oauth.get_song(config["uploads"]["private_upload_id"])  # private upload
         assert len(song) == 5


sigma67:
This is just how it works - if you're not authenticated you don't have access to the premium tracks, so they're replaced with music videos. This is a server side issue, not really something you can fix on the client side.

kaneryu:
No, this fixes it, i've been actively using this fix in my music player and it works as intended (logged out)

ytmusicapi/mixins/browsing.py
```diff
@@ -568,8 +574,23 @@ def get_album(self, browseId: str) -> JsonDict:
...
         results = nav(response, [*TWO_COLUMN_RENDERER, "secondaryContents", *SECTION_LIST_ITEM, *MUSIC_SHELF])
-        album["tracks"] = parse_playlist_items(results["contents"], is_album=True)
-
+        
+        if not clean_songs:
+            album["tracks"] = parse_playlist_items(results["contents"], is_album=True)
+        else:
+            urlCanonical: str | None = nav(response, ["microformat", "microformatDataRenderer", "urlCanonical"], True)
+            if urlCanonical:
+                playlistid = urlCanonical.split("list=")[-1]
```
sgvictorino: Does this value differ from `audioPlaylistId`?

https://github.com/sigma67/ytmusicapi/blob/f90e616c1a7da8f0b60eadb5b63d3a2ed8598088/ytmusicapi/parsers/albums.py#L33

kaneryu: Thanks for the tip! was able to simplify the code a bit

sigma67:
> No, this fixes it, i've been actively using this fix in my music player and it works as intended (logged out)

Sure does, but you can achieve the exact same result application side by calling `get_watch_playlist` yourself.

I want to avoid adding unnecessary maintenance burden by implementing "convenience" functionality in this library, where the same result can be achieved by simply looping over the videoIds returned by `get_album`.

Also you can use the audio playlist, as @sbaum  noted.

kaneryu:
Fair, alright

Issue #769, opened 2025-05-05
Support new monthly audience metric


sigma67:
**Is your feature request related to a problem? Please describe.**

Popular channels replace subscribers with monthly audience metric

https://music.youtube.com/channel/UCqJnSdHjKtfsrHi9aI-9d3g

**Describe the solution you'd like**

Parse monthly audience metric correctly into separate attribute `monthlyListens` and set `subscribers` to None as it is replaced in that case

**Describe alternatives you've considered**

-

**Additional context**

- https://support.google.com/youtubemusic/answer/15621827?visit_id=638820282146302918-1351610219&p=ytmMonthlyListens&rd=1
- https://9to5google.com/2025/01/07/youtube-music-monthly-audience-metric/

```
To display Monthly Audience your channel must meet requirements:

    Your channel is an [Official Artist Channel](https://support.google.com/youtube/answer/7336634) (OAC).
    Your channel has more than 100,000 subscribers.
    Your channel’s Monthly Audience is over 50,000 listeners.

If your channel does not meet these requirements, it will continue to only display the Subscriber count.
```

PR #849, opened 2025-12-22
feat(artists): add new monthly audience feature to get_artist (#769)

kaneryu:


ytmusicapi/mixins/browsing.py
```diff
@@ -267,6 +268,12 @@ def get_artist(self, channelId: str) -> JsonDict:
         artist["shuffleId"] = nav(header, ["playButton", "buttonRenderer", *NAVIGATION_PLAYLIST_ID], True)
         artist["radioId"] = nav(header, ["startRadioButton", "buttonRenderer", *NAVIGATION_PLAYLIST_ID], True)
         artist["subscribers"] = nav(subscription_button, ["subscriberCountText", "runs", 0, "text"], True)
+        artist["monthlyListeners"] = nav(header, ["monthlyListenerCount", "runs", 0, "text"], True)
+        artist["monthlyListeners"] = (
+            artist["monthlyListeners"].replace(" monthly audience", "")
+            if artist["monthlyListeners"]
```
sigma67: That's not how dictionaries work at all. You'll get a KeyError. Use the nav function instead of this if construct

kaneryu: Which part is wrong? monthlyListeners is always `str | None`, so if it's a str we replace, if None we ignore it. The `nav` above ensures this behavior

ytmusicapi/mixins/browsing.py
```diff
@@ -177,6 +177,7 @@ def get_artist(self, channelId: str) -> JsonDict:
                 "shuffleId": "RDAOkjHYJjL1a3xspEyVkhHAsg",
                 "radioId": "RDEMkjHYJjL1a3xspEyVkhHAsg",
                 "subscribers": "3.86M",
+                "monthlyListeners": "29.1M",
```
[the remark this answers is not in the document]
kaneryu: fixed

PR #854, opened 2026-01-21
fix: handle OLAK chart playlists without album info

### LINKED ISSUE #853: get_playlist() crashes on OLAK chart playlists without album info
## Description

`get_playlist()` crashes with `TypeError: 'NoneType' object is not subscriptable` when fetching OLAK playlists that don't have album info on all tracks (e.g., chart/trending playlists).

## Reproduction

**Library version:** 1.11.5

**Steps:**
```python
ytm = YTMusic()
playlist = ytm.get_playlist("OLAK5uy_mzYnlaHgFOvLaxqIPnnouEr-idiUn4NIM")  # Trending 20 Spain
```

**Error:**
```
TypeError: 'NoneType' object is not subscriptable
  File "ytmusicapi/parsers/playlists.py", line 122, in parse_audio_playlist
    playlist["title"] = playlist["tracks"][0]["album"]["name"]
```

## Root Cause

`parse_audio_playlist()` assumes all OLAK playlists are album playlists where every track has album info. However, chart playlists like "Trending 20 Spain" use the OLAK prefix but contain music videos (OMVs) without album info.

## Proposed Fix

Make `parse_audio_playlist()` return `None` when not all tracks have album info, allowing the caller to fall back to regular playlist parsing.

I have a fix ready and will submit a PR.

guillevc:
Fixes #853

## Summary

Some OLAK playlists (like charts/trending playlists) contain music videos without album info. Previously, `parse_audio_playlist()` would crash trying to access `album["name"]` on tracks without album info.

**Example:** `OLAK5uy_mzYnlaHgFOvLaxqIPnnouEr-idiUn4NIM` ("Trending 20 Spain") contains OMVs where most tracks have `album: None`.

## Changes

- `parse_audio_playlist()` now returns `None` if not all tracks have album info, signaling to the caller to fall back to regular playlist parsing
- Simplified the routing logic in `get_playlist()` to check return value
- Added test case for chart playlist (`OLAK5uy_mzYnlaHgFOvLaxqIPnnouEr-idiUn4NIM`)

## Checklist

- [x] Pre-commit checks pass (`ruff`, `ruff format`, `mypy`)
- [x] Added test case with mock data
- [x] Existing tests pass


sigma67:
Could you please merge main to check if the tests now pass. I cannot edit this PR

ytmusicapi/mixins/playlists.py
```diff
@@ -139,7 +139,12 @@ def get_playlist(
 
         request_func_continuations: RequestFuncBodyType = lambda body: self._send_request(endpoint, body)
         if playlistId.startswith("OLA") or playlistId.startswith("VLOLA"):
-            return parse_audio_playlist(response, limit, request_func_continuations)
+            # Try parsing as audio/album playlist - returns None if not all tracks have album info
```
sigma67: Honestly it's a bit strange how you solved this imo. You need a better way to determine if the playlist is affected (structurally missing album info).

Wouldn't it be simpler to simply not skip this branch for `OLAK` playlists by updating the if condition?

Checking for the absence of some parsed items in the result of the parser is exceptionally fragile and could have any number of other causes (empty playlist for example)

sigma67 (changes requested):

ytmusicapi/mixins/playlists.py
```diff
@@ -138,7 +138,9 @@ def get_playlist(
         response = request_func("")
 
         request_func_continuations: RequestFuncBodyType = lambda body: self._send_request(endpoint, body)
-        if playlistId.startswith("OLA") or playlistId.startswith("VLOLA"):
+        is_olak = playlistId.startswith("OLA") or playlistId.startswith("VLOLA")
```
sigma67: Is this intentional? The variable is called "olak", but the contains check for "ola"

guillevc: Hi! Just renamed it to `is_olak` to match the actual `startsWith` checks. 

Cheers!

sigma67:
@sgvictorino could you have a look as well?

sigma67 (approved):

sgvictorino (approved):
LGTM!

Issue #855, opened 2026-01-24
UnicodeError when headers contain unicode


mokurin000:
- [x] I confirm that I have read the [FAQ](https://ytmusicapi.readthedocs.io/en/stable/faq.html#why-is-ytmusicapi-returning-more-results-than-requested-with-the-limit-parameter)

**Describe the bug**

I just find the `__Secure-3PSID=` in Cookies may contains Unicode, and since ytmusicapi does not process URLEncode for cookies, it may finally result:

```text
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\ytmusicapi\mixins\watch.py", line 141, in get_watch_playlist   
    response = self._send_request(endpoint, body)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\ytmusicapi\ytmusic.py", line 231, in _send_request
    response = self._session.post(
               ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\requests\sessions.py", line 637, in post
    return self.request("POST", url, data=data, json=json, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\requests\sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\requests\sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\requests\adapters.py", line 644, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
  File "C:\Users\moku\Repos\FeelUOwn\.venv\Lib\site-packages\urllib3\connectionpool.py", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
[...]

PR #856, opened 2026-01-24
fix: UnicodeError for non-ascii headers

mokurin000:
solve #855 

## Changes

All file open are forced to use 'utf-8' encoding, regardless on what platform, on python it was likely utf-8 on Linux and MacOS, and ANSI for Windows.

Also, this ensures headers to be encoded in URLEncode, thus, unicode characters are encoded as `%E7%E6%F7`, e.g.

```python
from string import printable
from urllib.parse import quote

# '__Sec=%E4%BD%A0%E5%A5%BD,Hello; __Sec1=%E4%B8%96%E7%95%8C,World'
quote("__Sec=你好,Hello; __Sec1=世界,World", safe=printable)
```

### DIFF
--- a/tests/mixins/test_library.py
+++ b/tests/mixins/test_library.py
@@ -97,7 +97,11 @@ class TestLibrary:
         songs = yt_oauth.get_history()
         assert len(songs) > 0
         assert all(song["feedbackToken"] is not None for song in songs)
-        assert all(song["listenAgainFeedbackTokens"] is not None for song in songs)
+        assert all(
+            song["listenAgainFeedbackTokens"] is not None
+            for song in songs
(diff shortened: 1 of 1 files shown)


sigma67:
Another test is failing now: `test_setup_oauth`

mokurin000:
> Another test is failing now: `test_setup_oauth`

Fixed now, `NamedTemporaryFile` opens `w+b` by default, now it's set to w+.

sigma67:
Yeah it's some German TV Show, so it might not be available to some runners in the US West region. Unfortunately I don't know of a way to pick the region of the Github runner, seems random (and I don't want to use a proxy)

PR #865, opened 2026-01-31
fix browser setup for Chrome (#857)

### LINKED ISSUE #857: "Authentication Failed" when pasting request headers from Chrome

**Describe the bug**

When pasting request headers from Chrome (even though Firefox is recommended), the authentication process fails with:
```
ytmusicapi.exceptions.YTMusicUserError: The following entries are missing in your headers: x-goog-authuser. Please try a different request (such as /browse) and make sure you are logged in.
```

The reason this happens is because starting sometime early last year, Chrome request headers are copied to your clipboard without a colon at the end of each key. For example:
```
accept
*/*
accept-encoding
gzip, deflate, br, zstd
accept-language
en-US,en;q=0.9
```

This breaks the parser in `browser.py:setup_browser`.

I've been working on a solution but it's a bit tricky since now the header keys and header values don't have any distinguishing characteristics from each other -- you kind of just have to assume the first line is a key and the 2nd line is a value and go from there.

References:
* Closed ticket from 2023 that fixed Chrome headers when they used to have colons: #409 
* Discussion on this I opened last year: #793 
* Fix I implemented in my own project last year before I knew about issue 409 : https://github.com/apastel/ytmusic-deleter/blob/56726397b849d523d16de1e8799036e7a6133b4c/ytmusic_deleter/common.py#L94

We _could_ port over that fix to ytmusicapi but I'm not super proud of it. I would rather start with the existing fix that uses `chrome_remembered_key` and modify that instead. I started to work on something but so far it's not quite working yet.

**ytmusicapi version**

1.11.4

**To Reproduce**

1. Follow instructions to paste request headers, use Chrome instead of Firefox
2. Observe error in terminal.

apastel:
Fixes #857 

Supports all three of the following formats for pasting headers:

1. Firefox
2. Chrome (new format without colons)
3. Chrome (old format with colons on header keys)

Had to add a corner case to handled the odd `Decoded:` header in Chrome, which looks like the following:
```
Decoded:
message ClientVariations {
  // Active Google-visible variation IDs on this client. These are reported for analysis, but do not directly affect any server-side behavior.
  repeated int32 variation_id = [3300102, 3300132, 3313321, 3323055, 3330197, 3362821, 3393046, 3393674, 3395745, 3396053, 3396249, 3396352, 3396365, 3396446, 3396569, 3396819, 3396954, 3397042, 3397046];
  // Active Google-visible variation IDs on this client that trigger server-side behavior. These are reported for analysis *and* directly affect server-side behavior.
  repeated int32 trigger_variation_id = [3392236, 3396349];
}
```
It's the only one that still has a colon on the header key so it had to be handled differently. The end result in the parsed `user_headers` is the same as it was before, where the property ends up being `Decoded: "}"`


apastel:
Anyone had time to test this out yet? Just checking.

apastel:
Please let me know if there's anything faulty or missing from this PR and I will update. Thanks.

sigma67:
Hi, thanks for the contribution! Could you add a test that asserts that it's working as expected?

apastel:
> Hi, thanks for the contribution! Could you add a test that asserts that it's working as expected?

### Test Steps
1. Paste raw headers from Chrome into `headers_raw` in `test.cfg` (overflowing lines must be indented)
2. Run `pdm run pytest -k test_setup_browser`
3. Observe test passes (on `main`, this fails)
4. Re-run steps 1-3 with raw headers from Firefox to ensure there are no regressions.

sigma67:
I'm looking for an automated test (pytest), thanks

So basically add headers in the new/supported formats as a static file, while redacting confidential information

No manual steps should be needed

apastel:
> I'm looking for an automated test (pytest), thanks
> 
> So basically add headers in the new/supported formats as a static file, while redacting confidential information
> 
> No manual steps should be needed

Done.  
Let me know if I should delete the existing `test_setup_browser` along with the `headers_raw` config item in `test.example.cfg` since this new test covers everything that existing test did without the need for manual pasting.

apastel:
Any feedback on my pytest? Thanks.

sigma67:
Sorry I seem to have missed that. Thanks for adding the tests.

At first glance it seems that a lot of tests are failing. Will have a look at the code later today.

apastel:
> At first glance it seems that a lot of tests are failing. Will have a look at the code later today.

Ok, I better have another look at that too. I mistakenly thought it was the code coverage check failing to generate a report for some external reason, but upon closer look I see tests erroring out:

```
TypeError: RefreshingToken.__init__() got an unexpected keyword argument 'accept'
```
```
ytmusicapi.exceptions.YTMusicUserError: oauth JSON provided via auth argument, but oauth_credentials not provided.Please provide oauth_credentials as specified in the OAuth setup documentation.
```

apastel:
Fixed issue with failing tests. It was because my new test was creating a new `browser.json` with [REDACTED] values and this was messing up all the subsequent tests.

Now, there are two unrelated tests failing, both referenced in this other developer's PR from a couple days ago: https://github.com/sigma67/ytmusicapi/pull/884#issue-4073755058

apastel:
@sigma67 is the pipeline in a weird state currently with regard to secrets? I saw your comment on another PR and wondering if that's why the sudden errors with `config[auth]`: https://github.com/sigma67/ytmusicapi/pull/895#issuecomment-4170221506

sigma67:
@apastel yes, I had removed the secrets until we can get to an approval-only workflow. I will re-add them soon now that we've figured something out.

sigma67 (approved):

PR #866, opened 2026-02-01
Fix for #711: Implement custom playlist thumbnail uploads

### LINKED ISSUE #711: ability to add a playlist image via create_playlist()
Seems like youtube music recently added the ability to add custom images to playlists! Can we extend `create_playlist()` and `edit_playlist` to take an optional argument (an image file) so that we can programmatically set it?

Daveson217:
### Closes #711 
This PR adds the ability to upload custom thumbnails to playlists via edit_playlist.

_**Changes:**_
1. Implemented a new helper method _upload_playlist_thumbnail in PlaylistsMixin. This replicates the resumable upload protocol (handshake + binary upload) used elsewhere in the codebase for song uploads, but adapted for the playlist image endpoint.
2. Updated edit_playlist to accept a thumbnail argument (file path).
3. Added logic to upload the image first, retrieve the encryptedBlobId, and append the ACTION_SET_CUSTOM_THUMBNAIL action to the request body.

_**Testing:**_
1. Verified locally by uploading a JPG image to a personal playlist.
2. Confirmed the thumbnail updated successfully on the YouTube Music web interface.

_**Test used:**_
`yt.edit_playlist(playlistId="...", thumbnail="cover.jpg")`

P.S.: I am happy to add a unit test for `_upload_playlist_thumbnail` if you could point me to an example of how you prefer to mock upload requests.


ytmusicapi/mixins/playlists.py
```diff
@@ -430,3 +451,52 @@ def remove_playlist_items(self, playlistId: str, videos: JsonList) -> str | Json
...
+            raise YTMusicUserError(f"File {filepath} does not exist")         
+                
+        filesize = Path(filepath).stat().st_size
+        if filesize >= 314572800:  # 300MB in bytes
+            msg = f"File {filepath} has size {filesize} bytes, which is larger than the limit of 300MB"
+            raise YTMusicUserError(msg)
+        
+        headers = self.headers.copy()  
+        # Get upload_url by sending an empty request to the upload endpoint      
+        upload_url = "https://music.youtube.com/playlist_image_upload/playlist_custom_thumbnail"
```
sigma67: Please use the constant here: https://github.com/sigma67/ytmusicapi/blob/main/ytmusicapi/constants.py#L1

Daveson217:
@sigma67 here's what I have done so far:

1. I added a test_edit_playlist_thumbnail function

2. Refactored `upload_url` to use `YTM_DOMAIN`

3. Refactored `edit_playlist` method definition parameter, thumbnail, to be a `Path` from `pathlib`.

4. Added a test image in tests/data/mock_playlist_image.jpg

5. Changed Content-Type header for thumbnail binary upload. The `[music.youtube.com](http://music.youtube.com/)` endpoint seems to reject (400 Bad Request error) binary uploads if the `Content-Type` is set to `application/x-www-form-urlencoded`. I then used mimetypes to determine the type of the image (e.g., image/jpeg).

Lastly, YoutubeMusic requires an image with a minimum dimension of 360x360. I couldn't find a straightforward way to implement image size checking, so I didn't add this check.

For the most recent build, I had tested the `test_playlists.py` file, and it passed locally but failed here. I will try to find out what's happening.

Daveson217:
@sigma67  I fixed the mypy and ruff issues.
The tests pass on my end locally. 
Testing with `yt.edit_playlist(playlistId="...", thumbnail="cover.jpg")` also works fine locally.

What do you suggest?

sigma67:
I'm getting the same error locally, so not sure how I can help you here

```
Server returned HTTP 403: Forbidden.
```

sigma67:
If you suspect it's related to the brand account, you can also try using a different fixture like yt_auth. Although in the end it should also work with the brand account

Daveson217:
@sigma67 Thanks. I tried yt_oauth, but I ran into another error (build [9d2a842](https://github.com/sigma67/ytmusicapi/pull/866/commits/9d2a842c3b94f181f0df5bb0d0db2041a1162d98)).

I am now trying to backtrack to see why it works fine on my PC but fails on CI and locally at your end.

I am trying to understand more: for the CI tests and prolly locally at your end, do you use oauth or browser auth?

Daveson217:
I now recall something else @sigma67 . To be able to add custom playlist thumbnails at all, you need to first go to the playlist and enable the function by verifying your phone number.

Do you think this could be the cause of the 403 error?

```
 ytmusicapi.exceptions.YTMusicServerError: Server returned HTTP 403: Forbidden.
 The caller does not have permission
```

sigma67:
My account has a verified phone number. I can change the image just fine on the web UI.

sigma67:
You should double check that you are using the same account as with the `edit_playlist` test - there are multiple accounts here, only the one that created it will be able to edit it.

sigma67 (changes requested):
Unfortunately the new test still fails

PR #876, opened 2026-02-17
Support votes edit playlist

aishayacoob:
Adds support for playlist item voting in edit_playlist by introducing
an optional `voting` parameter (int | None).

Implements ACTION_SET_ALLOW_ITEM_VOTE with itemVotePermission
based on observed YouTube Music web payload.

The parameter is optional, preserving backward compatibility.


ytmusicapi/mixins/playlists.py
```diff
@@ -288,6 +288,7 @@ def edit_playlist(
         moveItem: str | tuple[str, str] | None = None,
         addPlaylistId: str | None = None,
         addToTop: bool | None = None,
+        voting: int | None = None,
```
sigma67, line 291: Why an integer? How would I know what to pass here as a user?

sigma67:
- a test is missing
- please make it transparent if you have used AI to create this PR

PR #879, opened 2026-02-23
edit_playlist: support collaboration

sgvictorino:


ytmusicapi/mixins/playlists.py
```diff
@@ -281,12 +284,33 @@ def create_playlist(
         response = self._send_request(endpoint, body)
         return response["playlistId"] if "playlistId" in response else response
 
+    def join_collaborative_playlist(self, playlistId: str, joinCollaborationToken: str) -> str | JsonDict:
```
sigma67: If this is a  new endpoint it needs more documentation (docs, README)

sigma67:
I have some concerns regarding the return value when other parameters for edit_playlist are set at the same time as collaboration=true. It seems the results of the other inputs are ignored and only the collaboration value is returned due to the way the return statement is set up.

However I'll merge it for now, feel free to fix this as you see fit.

sigma67 (approved):

sgvictorino:
> I have some concerns regarding the return value when other parameters for edit_playlist are set at the same time as collaboration=true. It seems the results of the other inputs are ignored and only the collaboration value is returned due to the way the return statement is set up.

https://github.com/sigma67/ytmusicapi/blob/416b5d9ce6924b113f49e36fc3b2f16d6228fb33/ytmusicapi/mixins/playlists.py#L383-L387

We only return early for `"STATUS_SUCCEEDED"`, which is all that would've been returned otherwise:

https://github.com/sigma67/ytmusicapi/blob/416b5d9ce6924b113f49e36fc3b2f16d6228fb33/ytmusicapi/mixins/playlists.py#L390

Let me know if I misunderstood something.

PR #893, opened 2026-03-29
fix: browser auth detection and unexpected OAuth token fields

### LINKED ISSUE #887: OAuth device flow fails due to unexpected `refresh_token_expires_in` field
* [x] I confirm that I have read the [[FAQ](https://ytmusicapi.readthedocs.io/en/stable/faq.html#why-is-ytmusicapi-returning-more-results-than-requested-with-the-limit-parameter)](https://ytmusicapi.readthedocs.io/en/stable/faq.html#why-is-ytmusicapi-returning-more-results-than-requested-with-the-limit-parameter)

**Describe the bug**

Running `ytmusicapi oauth` (device flow) fails after completing Google authentication with a `TypeError`. The error occurs because the OAuth response now includes a new field `refresh_token_expires_in`, which is not expected by `RefreshingToken.__init__()`.

---

**ytmusicapi version**

1.11.5

---

**To Reproduce**

1. Run `ytmusicapi oauth`
2. Enter valid Google OAuth client ID and secret
3. Open the provided device verification URL and complete login
4. Return to terminal and press Enter
5. Observe crash

Error:

```
TypeError: RefreshingToken.__init__() got an unexpected keyword argument 'refresh_token_expires_in'
```

---

**Additional context**

It appears that Google recently added the `refresh_token_expires_in` field to the OAuth device flow response.

The issue occurs at `ytmusicapi\auth\oauth\token.py`, line 131, in `prompt_for_token`:

```python
ref_token = cls(credentials=credentials, **raw_token)
```

A potential fix would be to ignore unknown fields before passing them:

```python
raw_token.pop("refresh_token_expires_in", None)
ref_token = cls(credentials=credentials, **raw_token)
```

Is handling it this way sufficient, or would you prefer a more generic approach to ignore unexpected OAuth fields?

Happy to open a PR if needed.

hb69dpsgw4-source:
## Summary

Two fixes for authentication issues in ytmusicapi 1.11.5:

### 1. `determine_auth_type` misidentifies browser auth as `OAUTH_CUSTOM_CLIENT`

Browser auth files created via `ytmusicapi setup` contain cookies with SAPISID but no `authorization` header (it's computed at request time from the SAPISID cookie). The current implementation defaults to `OAUTH_CUSTOM_CLIENT` when no `authorization` header is present, which causes the constructor to require `oauth_credentials` and reject valid browser auth files.

**Fix:** Check for SAPISID in cookies as a fallback when neither an `authorization` header nor OAuth token fields are found. Also restructured the function to use early returns for clarity.

### 2. Google's OAuth response now includes `refresh_token_expires_in` (Fixes #887)

Google recently added a `refresh_token_expires_in` field to the OAuth device flow token response. Since `Token` is a strict dataclass, passing this unexpected field causes:

```
TypeError: RefreshingToken.__init__() got an unexpected keyword argument 'refresh_token_expires_in'
```

**Fix:** Filter to recognized `Token` fields before constructing the token object, both in:
- `RefreshingToken.prompt_for_token()` — when creating new tokens from the OAuth flow
- `OAuthToken.from_json()` — when loading previously saved tokens that may contain the extra field

This approach is forward-compatible: any future fields Google adds to the response will be safely ignored.

## Testing

- Verified browser auth detection works correctly with SAPISID cookies
- Verified OAuth device flow completes without `refresh_token_expires_in` error
- Verified previously saved token files with extra fields load correctly

## Changes

- `ytmusicapi/auth/auth_parse.py` — restructured `determine_auth_type` with SAPISID cookie detection
- `ytmusicapi/auth/oauth/token.py` — filtered unknown fields in `prompt_for_token` and `from_json`

### DIFF
--- a/ytmusicapi/auth/auth_parse.py
+++ b/ytmusicapi/auth/auth_parse.py
@@ -55,14 +55,19 @@ def determine_auth_type(auth_headers: CaseInsensitiveDict[str]) -> AuthType:
     :param auth_headers: auth headers dict
     :return: AuthType enum
     """
-    auth_type = AuthType.OAUTH_CUSTOM_CLIENT
-    if OAuthToken.is_oauth(auth_headers):
-        auth_type = AuthType.OAUTH_CUSTOM_CLIENT
-
     if authorization := auth_headers.get("authorization"):
         if "SAPISIDHASH" in authorization:
-            auth_type = AuthType.BROWSER
-        elif authorization.startswith("Bearer"):
-            auth_type = AuthType.OAUTH_CUSTOM_FULL
+            return AuthType.BROWSER
+        if authorization.startswith("Bearer"):
+            return AuthType.OAUTH_CUSTOM_FULL
 
-    return auth_type
+    if OAuthToken.is_oauth(auth_headers):
+        return AuthType.OAUTH_CUSTOM_CLIENT
+
+    # Browser auth files contain cookies with SAPISID but no authorization
+    # header (it's computed at request time from the SAPISID cookie).
+    cookie = auth_headers.get("cookie", "")
+    if "SAPISID" in cookie:
+        return AuthType.BROWSER
+
+    return AuthType.OAUTH_CUSTOM_CLIENT
--- a/ytmusicapi/auth/oauth/token.py
+++ b/ytmusicapi/auth/oauth/token.py
@@ -74,6 +74,10 @@ class OAuthToken(Token):
             with open(file_path, encoding="utf-8") as json_file:
                 file_pack = json.load(json_file)
 
+        # Filter to recognized Token fields — previously saved tokens may
+        # contain extra fields from Google's OAuth response.
+        known_fields = set(Token.members())
+        file_pack = {k: v for k, v in file_pack.items() if k in known_fields}
         return cls(**file_pack)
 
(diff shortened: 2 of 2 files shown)


sigma67 (changes requested):
Hi @hb69dpsgw4-source , unfortunately this PR contains no tests and the checks are failing

sigma67:
Definitely open to merging this though if you want to circle back around to finish it @hb69dpsgw4-source  :)

Wolf1098:
I suspect this a bypass to the underlying issue... I suspect and believe that the `refresh_token_expires_in` is not to be filtered, but parsed, and used to refresh the refreshing_token(that is used to refresh its oauth api token 'token'?)... this should, in my theory, allow the client to know when its permission to ask google for a new "oauth-api(ytdata)" `token` is about to expire, and when it should go refresh its `refreshing_token` that is used to ask google for that `token`

imthenachoman:
When will this fix be applied? I'm unable to `ytmusicapi oauth` because I get the `TypeError: RefreshingToken.__init__() got an unexpected keyword argument 'refresh_token_expires_in'` error.

sigma67:
Since there is no more activity here I will close this in favor of the other PR

PR #895, opened 2026-04-01
fix - parse all of artist description runs instead of just the first one

### LINKED ISSUE #877: Descriptions cut off where links are

**Describe the bug**

When getting the description of an artist, it seems to only pull the plain text. Once it hits a link it will cut the description off there

**ytmusicapi version**

1.11.5

**To Reproduce**

1. Use YTMusic.get_artist to get details about Hatsune Miku
2. Description ends with "From Wikipedia ("

**Additional context**

I've encountered this problem when making the artist page for https://github.com/m-obeid/Muse

ThomasNguyen5000:
- Fix parsing artist description by concatenating all the description's runs instead of just taking the first one
- Also format hyperlink in the description in markdown format.
- Added tests for the new feature

Fix bug mentioned in #877


sigma67:
Hi, thanks for your contributions. I'll try to get the pipeline running again later today so we can get things moving.

Someone tried to steal secrets so I had to lock it down a bit

ytmusicapi/helpers.py
```diff
@@ -98,3 +100,25 @@ def sum_total_duration(item: JsonDict) -> int:
...
+    if not isinstance(descriptionRunsList, list):
+        return ""
+
+    description = ""
+    for run in descriptionRunsList:
+        if "navigationEndpoint" in run:
+            link = nav(run, ["navigationEndpoint", "urlEndpoint", "url"])
+            desc = run["text"]
+
+            description += format_markdown_link(desc=desc, link=link)
```
sigma67: We shouldn't return markdown here. This forces the user to parse markdown if they need the link in a different format (like RST or href or only the link). Instead, return both separately? Maybe you have a different idea

ThomasNguyen5000: I reckon we can instead of just returning text, we can return a list of dictionaries which represent the runs themselves.

ThomasNguyen5000: I have modified the returned description to be a list of runs instead of just plain text. Lmk what you think.

ThomasNguyen5000: I thought of a new solution to ensure backwards compatibility. We can split the hyperlinks into a new field e.g. `descriptionHyperLink` and that field will have all the positions of the hyperlinks and where they are linking to, e.g. the shape will be roughly

```
{ start: int, end: int, url: str }
```
Where start and end will be the index of the starting point of the hyperlink string and similarly for end. What do you think?

Will work on it when I am free next

sigma67: This issue is definitely more difficult than it looks.

I like the current state, but the problem is that it's a breaking change. It changes the returned format of `description` from string to dictionary.

So I would propose the following:

- return `description` in full, using only the `text` attributes like the YTM visible description
- additionally return `descriptionRuns`, which provides access to the original run format (i.e. the currently implemented state for `description`)

ThomasNguyen5000:
@sigma67  it seems that the new changes in the PR #920 have broken the ruff action

sigma67:
A test is still failing and seems it's not flaky

sgvictorino:
These assertions need to account for the addition of `descriptionRuns`:

```diff
--- a/tests/mixins/test_browsing.py
+++ b/tests/mixins/test_browsing.py
@@ -32,7 +32,7 @@ class TestBrowsing:

     def test_get_artist(self, yt):
         results = yt.get_artist("MPLAUCmMUZbaYdNH0bEd1PAlAqsA")
-        assert len(results) == 17
+        assert len(results) == 18
         assert results["shuffleId"] is not None
         assert results["radioId"] is not None

@@ -148,7 +148,7 @@ class TestBrowsing:
             album = yt.get_album("MPREabc")
             assert len(album["tracks"]) == 19
             assert len(album["artists"]) == 1
-            assert len(album) == 14
+            assert len(album) == 15
             for track in album["tracks"]:
                 assert isinstance(track["title"], str) and track["title"]
                 assert len(track["artists"]) > 0
```

sigma67:
just to clarify, the fix you did is not related to your actual change in this PR?

ThomasNguyen5000:
@sigma67 Yeah it is not related. I suppose I should open a new issue/PR for that instead but I don't think this PR can pass the tests in the meanwhile

sigma67:
Ok we can merge it for now, feel free to follow up on it separately

PR #903, opened 2026-04-21
harden coverage.yml

sigma67:


sigma67:
@sgvictorino new approach looks promising, let me know what you think

image

sigma67:
from https://github.com/sigma67/ytmusicapi/pull/904

sigma67:
had to merge this first because the workflow file is loaded from the base branch (main) and otherwise the changes had no effect

.github/workflows/coverage-pr.yml
```diff
@@ -0,0 +1,62 @@
...
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    # Requires maintainer approval via the 'coverage' environment before the job
+    # runs and secrets are exposed. Configure in Settings → Environments →
+    # coverage → Required reviewers. Enable "Prevent administrators from
+    # bypassing configured protection rules" to enforce this for admins too.
+    environment: coverage
+    permissions:
+      contents: read
```
sgvictorino, line 20: Sorry, I didn't think we needed `contents` on a public repo.

.github/workflows/coverage-pr.yml
```diff
@@ -0,0 +1,62 @@
...
+permissions: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    # Requires maintainer approval via the 'coverage' environment before the job
+    # runs and secrets are exposed. Configure in Settings → Environments →
+    # coverage → Required reviewers. Enable "Prevent administrators from
+    # bypassing configured protection rules" to enforce this for admins too.
+    environment: coverage
```
sgvictorino, line 18: Instead of having two workflows, can we switch the environment name for `push` events?

sgvictorino (commented):
This is great! Much more robust than labeling.

.github/workflows/coverage-pr.yml
sgvictorino: Like this:

```yaml
environment: ${{ case(github.event_name == 'push', null, 'coverage') }}
```

sigma67: Thanks for the tip, i'm applying it. also enabled CodeQL scans so that should hopefully catch any new actions issues

PR #914, opened 2026-04-25
get_song_credits: normalize section keys across locales

sgvictorino:
`YTMusic(language="fr").get_song_credits("MPTCJpKLGcWBJ5Y").keys()`:

```diff
-['interprète(s)', 'auteur(s)/compositeur(s)', 'producteur(s)', 'métadonnées_du_titre']
+['performed_by', 'written_by', 'produced_by', 'music_metadata_provided_by', 'other_sections']
```


ytmusicapi/mixins/browsing.py
```diff
@@ -635,11 +634,14 @@ def get_song_credits(self, browseId: str) -> JsonDict:
...
+        localized_section_map = self.parser.get_song_credit_section_map()
         for section in sections:
             section_data = section["dismissableDialogContentSectionRenderer"]
 
-            section_pretty_name: str = nav(section_data, TITLE_TEXT)
-            section_snake_case_name = section_pretty_name.lower().replace(" ", "_")
+            section_local_name: str = nav(section_data, TITLE_TEXT)
+            section_snake_case_name = localized_section_map.get(section_local_name)
+            if section_snake_case_name is None:  # pragma: no cover
+                continue
```
sgvictorino: Should we preserve unknown section titles?

ytmusicapi/parsers/i18n.py
```diff
@@ -49,6 +49,15 @@ def get_api_result_types(self) -> list[str]:
             *self.get_search_result_types(),
         ]
 
+    @i18n
+    def get_song_credit_section_map(self) -> JsonDict:
+        return {
```
sigma67, line 54: This can be simplified to avoid the string repetitions

```py
@i18n
def get_song_credit_section_map(self) -> JsonDict:
    SECTION_KEYS = (
        "performed_by",
        "written_by",
        "produced_by",
        "music_metadata_provided_by",
    )
    return {_(key): key for key in SECTION_KEYS}
```

ytmusicapi/mixins/browsing.py
```diff
@@ -635,11 +634,14 @@ def get_song_credits(self, browseId: str) -> JsonDict:
 
         credits: JsonDict = {}
         sections = nav(response, CREDITS_SECTIONS)
+        localized_section_map = self.parser.get_song_credit_section_map()
         for section in sections:
             section_data = section["dismissableDialogContentSectionRenderer"]
 
-            section_pretty_name: str = nav(section_data, TITLE_TEXT)
-            section_snake_case_name = section_pretty_name.lower().replace(" ", "_")
```
sigma67, line 642: any reason why you removed this?

sgvictorino: - I renamed `section_pretty_name` to the more descriptive `section_local_name`
- `section_snake_case_name` is now the result of the `localized_section_map` lookup (e.g. "music_metadata_provided_by") instead of `section_local_name.lower().replace(" ", "_")` ("Métadonnées du titre" -> "métadonnées_du_titre")

ytmusicapi/parsers/i18n.py
sgvictorino: I'm fine with that but wouldn't those be the first strings in the file that aren't picked up by `xgettext`?

sigma67: right I guess that would break the automated workflow described in the locales readme

ytmusicapi/mixins/browsing.py
sigma67: If we continue here doesn't this mean that we lose information compared to before? i.e. we don't return information that was retrieved from the server and was returned to the user previously?

Without this change the fact known/unknown didn't matter, it just always used the localized snake cased name

sigma67: So I guess the best we can do for those cases is to keep the previous behavior.

sgvictorino: Does snake casing them still make sense? It might be worth returning the unknown titles as-is for the sake of end users. Maybe nested under `localized_sections`?

```json
{
  "performed_by": [""],
  "localized_sections": [
    { "title": "Drum Programming", "data": [""] }
  ]
}
```

sigma67: I guess the benefit of having them nested separately in `localized_sections` is that you can easily iterate over those not known ahead of time in a different way. So yeah, it can do without the snake casing

sigma67: On the other hand, for those we are translating to English, we now lose the original completely, so I don't think anyone depends on the precise localized form

sigma67: Fine with me either way

sgvictorino: > On the other hand, for those we are translating to English, we now lose the original completely

Fair, how about this?

```json
{
  "performed_by": { "localized_title": "Performed by", "data": [""] },
  "other_sections": [
    { "localized_title": "Drum Programming", "data": [""] }
  ]
}
```

sigma67: Sounds good to me

PR #927, opened 2026-05-05
fix(oauth): Handle refresh token expiry field

### LINKED ISSUE #887: OAuth device flow fails due to unexpected `refresh_token_expires_in` field

vandit98:
## Summary

Hi, Vandit here. This PR makes OAuth token setup ignore unexpected fields returned by Google before constructing `RefreshingToken`. Google can now include `refresh_token_expires_in`, which caused `RefreshingToken.__init__()` to raise a `TypeError` during device flow setup.

Fixes #887.

## Verification

- `python3 -m pytest -o addopts="" tests/auth/test_oauth.py::TestOAuth::test_setup_oauth_ignores_unexpected_token_fields`
- `python3 -m ruff check ytmusicapi/auth/oauth/token.py tests/auth/test_oauth.py`
- `git diff --check`

Note: the existing `test_setup_oauth` could not run locally because this checkout does not include a populated `tests/test.cfg` auth fixture (`KeyError: auth`). Plain `mypy ytmusicapi/auth/oauth/token.py` also reports local baseline/tooling issues around missing `requests` stubs and unrelated files.

### DIFF
--- a/ytmusicapi/auth/oauth/token.py
+++ b/ytmusicapi/auth/oauth/token.py
@@ -4,7 +4,7 @@ import webbrowser
 from collections.abc import KeysView
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, cast
+from typing import Any
 
 from requests.structures import CaseInsensitiveDict
 
@@ -110,13 +110,6 @@ class RefreshingToken(OAuthToken):
         self._local_cache = path
         self.store_token()
 
-    @staticmethod
-    def _supported_token_fields(token: RefreshableTokenDict) -> RefreshableTokenDict:
-        return cast(
-            RefreshableTokenDict,
-            {key: value for key, value in token.items() if key in Token.members()},
-        )
-
     @classmethod
     def prompt_for_token(
         cls, credentials: OAuthCredentials, open_browser: bool = False, to_file: str | None = None
@@ -135,8 +128,16 @@ class RefreshingToken(OAuthToken):
             webbrowser.open(url)
         input(f"Go to {url} , finish the login flow and press Enter when done, Ctrl-C to abort")
         raw_token = credentials.token_from_code(code["device_code"])
-        ref_token = cls(credentials=credentials, **cls._supported_token_fields(raw_token))
-        ref_token.update(ref_token.as_dict())
+        refresh_token_expires_in = raw_token.get("refresh_token_expires_in", raw_token["expires_in"])
+        ref_token = cls(
+            credentials=credentials,
+            access_token=raw_token["access_token"],
+            refresh_token=raw_token["refresh_token"],
+            scope=raw_token["scope"],
+            token_type=raw_token["token_type"],
+            expires_in=refresh_token_expires_in,
+        )
+        ref_token.update(raw_token)
         if to_file:
             ref_token.local_cache = Path(to_file)
         return ref_token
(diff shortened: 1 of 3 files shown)


vandit98:
Hi, Vandit here. Following up on this PR since it has been open since May 6, 2026. I checked the failing `build` job in the Code coverage workflow, and the root cause appears to be an unrelated test failure in `tests/mixins/test_explore.py::TestExplore::test_get_explore` after retries. The subsequent Codecov "Found 0 coverage files" error is downstream from that, because `pdm run pytest --junitxml=test-results.xml` exits before `pdm run coverage xml` can produce `coverage.xml`. Recent `main` coverage runs are green, so this looks like a flaky live-data failure rather than something introduced by the OAuth-token-field change here. If a maintainer is able to rerun the failed job, I’d appreciate it, and I’m happy to adjust anything in the patch if you spot a code-level concern.

sigma67:
Hi @vandit98 , yes I'm on it. the test has been flaky for a while due to regional differences in data returned by YTM

ytmusicapi/auth/oauth/token.py
```diff
@@ -110,6 +110,13 @@ def local_cache(self, path: Path) -> None:
         self._local_cache = path
         self.store_token()
 
+    @staticmethod
+    def _supported_token_fields(token: RefreshableTokenDict) -> RefreshableTokenDict:
+        return cast(
+            RefreshableTokenDict,
+            {key: value for key, value in token.items() if key in Token.members()},
```
sigma67: cast is not the correct way to do this. you're sidestepping the fact that there is a new entry by working around the constructor.

I would prefer the solution from https://github.com/sigma67/ytmusicapi/pull/893/changes#diff-827f49453370105ce3f8675213ab23aafe980f8c22de6135dccfe84f36cb137aR77

or even better one that correctly recognizes the new `refresh_token_expires_in`, and uses it to replace the attribute `expires_in` we created for this purpose back when it was no provided by the server .

sigma67 (changes requested):

sigma67 (approved):
So this means that `refresh_token_expires_in` takes precedence over `expires_in`, which is fine I guess

imthenachoman:
I am still getting this error.

PR #939, opened 2026-06-05
default search: fix itemSection parsing

sigma67:


sigma67:
@sgvictorino somehow I forgot that you had already fixed this in the other PR. I tried accepting your fix as it seems more thorough than mine, but unfortunately it seemed to miss some cases (not all itemsections which don't contain actual results are "did you mean")

But we both independently arrived at essentially the same solution that's a good sign :)

Another thing I was not so sure about: Should we really parse the flexItems into category? Category was intended for the search sections back when they still existed, and the categories were plural. The first flexItem is used for the `result_type` value

I've removed that part as well now.

sgvictorino:
I agree the singular "categories" aren't useful.

PR #958, opened 2026-07-11
fix(parsers): parse localized view counts in non-English locales

### LINKED ISSUE #952: `YTMusic.search` method returns "views" wrongly (?) identified as an "artist" -...

**Describe the bug**

First of all, this only happen when the `YTMusic` instance was created using `langugae='ja'` and `location='JP'` or maybe for other non-English languages as well (?).

In the [ytmusicapi/parsers/songs.py](ytmusicapi/parsers/songs.py), the `runs` passed to `parse_song_runs()` usually contains following data:

```python
runs = [
    {"text": "Artist Name", "navigationEndpoint": {...}},
    {"text": " • "},
    # This is the views count (34.06 million views), but in Japanese as opposed to in English "34M views".
    {"text": "3406万回視聴"},
    {"text": " • "},
    {"text": "3:48"},
]
```

Then this function `parse_song_runs()` calls `parse_song_run()` with individual item as `run` from this `runs` list. And when it passes the `{"text": "3406万回視聴"}`, because it does not contain `"navigationEndpoint"`, the `else` block gets executed inside the `parse_song_run()` function. And the regex for matching the views never matches it as me think (not sure) it only checks for a string starting with number, followed by a single space and any character except space(?):

https://github.com/sigma67/ytmusicapi/blob/9ba806fa57414f0ad1054ac80074cb61bcbef9e4/ytmusicapi/parsers/songs.py#L32

And not other `elif` blocks match and it ends up in the `else` which marks it as an artist without id.

**ytmusicapi version**

1.12.1

**To Reproduce**

1. Initialize `YTMusic` class with `language='ja'` and `location="JP'`.
    - ```python
      ytmusic = YTMusic(
            language='ja',
            location='JP',
      )
      ```
3. Call the `search` method on that object with `query='Yorushika - Forget It'`
4. Check the returned result (maybe the first one - "Top Result" is enough).
5. The artist list contains two item, first one is correct artist 'ヨルシカ' and the second one is '3406万回視聴', which is not an artist (again, it means '34.06 million views').

**Additional context**

- Screenshot of the same search query on the YouTube Music website:
    - In the "Settings -> Language & location", "Location" set to "United States" and "Language" set to "English (US)":
    - <img width="864" height="139" alt="Image" src="https://github.com/user-attachments/assets/2fff9ece-e35e-41b0-9222-d4c5c5950f74" />
    - In the "Settings -> Language & location", "Location" set to "Japan" and "Language" set to "Japanese":
    - <img width="864" height="139" alt="Image" src="https://github.com/user-attachments/assets/40fc058b-6d91-48cb-af8c-4ca649dcdde1" />

;w; sorry for long text, and stupid explainations.. me tried to keep it as short as possible. thank you for your time and hard-work!!!!

Rishi943:
Fixes #952.

`parse_song_run()` classified a non-linked run as a view count only when it matched `^\d([^ ])* [^ ]*$` — i.e. it required an ASCII space between the number and the "views" word. CJK locales don't have one: with `language="ja"`, view counts arrive as e.g. `"3406万回視聴"`, so the run fell through to the `else` branch and was returned as a bogus unlinked artist alongside the real one.

Fix, keeping the existing heuristics' spirit:
- Check the fully-specific patterns first: duration (`(\d+:)*\d+:\d+`) and year (`^\d{4}$`) — unchanged.
- Then treat any remaining digit-leading run as views. Only split off the first token when a space is actually present; for scriptio-continua locales the full text is kept as the views value (there's no safe way to split the magnitude from the "views" word, and dropping it would lose information).

Digit-leading *unlinked artist* text was already matched by the old views regex in the common `"21 Savage"`-shaped case, so this doesn't introduce a new misclassification class — linked artist runs (with `navigationEndpoint`) are unaffected either way.

Verified against the runs from the issue (`ヨルシカ` + `3406万回視聴` + `3:48`):

```python
{'artists': [{'name': 'ヨルシカ', 'id': 'UC...'}], 'views': '3406万回視聴', 'duration': '3:48', 'duration_seconds': 228}
```

Regression-checked: `"34M views"` → `34M`, non-breaking-space variant, `"1,234,567 views"`, `"2020"` (year), `"3:48"`/`"1:23:45"` (duration), and the unlinked-artist fixture case `"The Wolfpack & Reel Wolf"` (still artist — doesn't start with a digit).

Note: written with AI assistance (Claude Code); verification outputs above were produced by running `parse_song_runs` directly on the issue's data.


sigma67:
I would honestly prefer to make the fix only for the affected language instead of changing the code path for all languages...

sigma67:
Thanks — right diagnosis, and it fixes more than the issue asked for. I verified against live YT Music responses in `en/de/ru/fr/es/it/ja/zh/ko/ar/hi` and pushed three commits to this branch; details below.

### Confirmed fixes
Besides ja/zh, the reorder also repairs locales whose views string has 3+ ASCII tokens, which `main` dropped into `artists`: `fr` `'1,7\xa0Md de vues'`, `es` `'2\xa0M de visualizaciones'`, `it` `'6,1\xa0Mln di visualizzazioni'`.

### 1. The `\xa0` note wasn't respected (19fd311)
The comment still said YT separates number and magnitude with `\xa0`, but the split was ASCII-space-only. `fr` uses `\xa0` for *both* separators, so nothing got stripped — one `fr` result page returned:

```
'1,7\xa0Md'   '6,1\xa0M'   '13\xa0M'                    # large counts: word stripped
'88\xa0k\xa0vues'  '202\xa0k\xa0vues'  '340\xa0vues'    # small counts: word kept
```

Now split on `\xa0` too and keep at most number + magnitude:

```python
head = text.split(" ")[0].split("\xa0")
return {"type": "views", "data": "\xa0".join(head[:2])}
```

Remaining known gap: `fr` `'340\xa0vues'` (no magnitude token) is indistinguishable from `de` `'1,7\xa0Mrd.'` without a locale word list, so it stays unsplit.

### 2. Bare `^\d` created a new misclassification class (f1eeffd)
Not just the `"21 Savage"` shape the description mentions — unlinked **single-token** digit-leading names (`2Pac`, `21`, `6ix9ine`) were artists on `main` and became view counts here. Added a guard: a view count always has a separator or non-ASCII text.

```python
elif re.match(r"^\d", text) and (" " in text or "\xa0" in text or not text.isascii()):
```

### 3. Tests (6b360ef)
`CONTRIBUTING.rst` asks for tests and `tests/parsers/test_songs.py` is offline-only, so added parametrized cases using verbatim API strings from all ten locales plus artist/duration/year negatives. All four of the new failing cases fail on 0f7fa4e and pass now.

### Not fixed: `ko` is still broken (same bug class as #952)
Korean puts the word first, so no digit-leading heuristic can reach it:

```
run:    '조회수 17억회'
result: views=None, artists=['Rick Astley', '조회수 17억회']
```

After this PR, 20/20 results parse views correctly in `en/de/ru/fr/es/it/ja/zh/ar/hi` and 0/20 in `ko`. Fixing that needs a positional rule (metadata run after duration) rather than a text heuristic, so I left it out of scope — #952 shouldn't be auto-closed by this merge.

sigma67:
Follow-up: `ko` is fixed too (896606f), and probing all 17 `SUPPORTED_LANGUAGES` turned up more of the same bug class than just Korean.

### What the API actually returns

Video results are number-first in most locales, but five languages put the word before the count — and song results ("plays") use a *different* shape again in the same locales:

| lang | videos | songs |
|---|---|---|
| ko | `조회수 17억회` | `3908만회 재생` |
| zh_TW | `觀看次數：17億次` | `播放次數：4505` |
| zh_CN | `17亿次观看` | `播放次数：3908万` |
| ja | `17億回視聴` | `再生回数 4.3億 回` |
| ur | `1.7\xa0ارب ملاحظات` | `\u202b3.5\xa0ارب بار چلائے گئے` (leading RTL mark, so not digit-leading either) |
| pt | `1,7\xa0bi de visualizações` | `Tocou 3,5\xa0bi vezes` |

### Fix

`parse_song_run`'s views logic moved into a `parse_views()` helper that first strips a leading views word, but **only for non-latin scripts**:

```python
VIEWS_PREFIX = re.compile(r"^\D*?[\s:\uff1a\u200e-\u200f\u202a-\u202e]")
```

`\D*?` cannot cross a digit, so at most a leading word (or bidi mark) is removed. Restricting it to non-latin text is what keeps `Take 5 Records` an artist — in latin script a prefixed count is indistinguishable from a name. Since a stripped prefix already proves the run is a count, the "bare ASCII token" guard is skipped in that case, which is what lets `播放次數：4505` → `4505` through while `2Pac` and `22/7` stay artists.

Also verified against `AKB48`, `乃木坂46`, `欅坂46`, `2NE1`, `℃-ute`, `Blink-182`, `Maroon 5`, `米津玄師` — all still artists (`乃木坂46` is the interesting one: no separator before the digits, so nothing is stripped).

### Result — live, all 17 languages, 20 video + 20 song results each

16/17 parse **every** count in both result types with no leak into `artists`. Before this PR the same probe missed 20/20 video counts in `fr/ja/es/it/zh/ko` and every song count in `ja/zh/ko/ur/pt`.

The one remaining gap is **`pt` song plays** (`Tocou 3,5\xa0bi vezes` → still an artist): latin script, word-first, and structurally identical to a band name like `Take 5 Records`. That one is pre-existing (not a regression from this PR) and can't be fixed by a text heuristic — it needs locale-aware parsing, e.g. a `views`/`plays` msgid in the gettext catalogs used via `Parser`. Probably worth its own issue.

Tests: `TestParseSongRun::test_views_with_leading_word` covers ko/zh_TW/zh_CN/ja/ur with verbatim API strings; all 7 cases fail on 6b360ef and pass now. Full file: 34 passed, ruff + mypy clean.

PR #959, opened 2026-07-19
fix: tolerate searchEndpoint (hashtag) runs in artist description

grray:
**Describe the bug**
Since #895, `get_artist()` raises `KeyError` for any artist whose description
contains a hashtag:

KeyError: "Unable to find 'urlEndpoint' using path ['navigationEndpoint',
'urlEndpoint', 'url'] on {'clickTrackingParams': '...', 'searchEndpoint':
{'query': '#1billboard', 'params': 'agIoAQ%3D%3D'}}, exception: 'urlEndpoint'"

**To Reproduce**
    YTMusic().get_artist("UCRHts_o65erNP48E0j2qYkw")  # Zacardi Cortez
    YTMusic().get_artist("UCxFCRsIS2KxYg6LB5OIc4Rw")  # Carolyn Arends

Both descriptions contain hashtags ("#1Billboard", "#canadianmusic").

**Cause**
`parse_description_runs()` assumes every run with a `navigationEndpoint`
contains an `urlEndpoint`, but hashtag runs contain a `searchEndpoint` instead.

**Fix**
Resolve the url with `nav(..., none_if_absent=True)` and fall back to a plain
text run when there is no url. Unit test included.


sigma67:
Could you please add a test that covers this case?

grray:
@sigma67 I've added the test. I hope I got everything right.

sigma67:
Thanks, it seems the failing tests are unrelated, I'll have to look at those separately first

PR #973, opened 2026-07-31
fix(parsers): tolerate playlist headers without secondSubtitle

### LINKED ISSUE #892: get_playlist method fails for some playlists

**Describe the bug**

`get_playlist` method fails for some playlists with `KeyError`.

**ytmusicapi version**

1.11.5

**To Reproduce**

The following code:
```py

YTMusic().get_playlist('PL8eCD0llEGnywNAITfRGFiCCNpmaHc1_l')
```
Results in the traceback:
```
  ...
    YTMusic().get_playlist('PL8eCD0llEGnywNAITfRGFiCCNpmaHc1_l')
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.14/site-packages/ytmusicapi/mixins/playlists.py", line 170, in get_playlist
    playlist.update(parse_playlist_header_meta(header))
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/app/.venv/lib/python3.14/site-packages/ytmusicapi/parsers/playlists.py", line 77, in parse_playlist_header_meta
    if "runs" in header["secondSubtitle"]:
                 ~~~~~~^^^^^^^^^^^^^^^^^^
KeyError: 'secondSubtitle'
```

Sanjays2402:
Closes #892

`parse_playlist_header_meta` indexed `header["secondSubtitle"]` directly, so a playlist whose header omits that field raised `KeyError: 'secondSubtitle'` and `get_playlist` failed outright. The sibling `title` lookup in the same function already guards with `.get`, so this does the same — `views`/`duration`/`trackCount` just stay `None`.

New `tests/parsers/test_playlists.py` fails with the original `KeyError` and passes with the fix.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

### DIFF
--- a/tests/parsers/test_playlists.py
+++ b/tests/parsers/test_playlists.py
@@ -1,0 +1,31 @@
+from ytmusicapi.parsers.playlists import parse_playlist_header_meta
+
+
+def _header(**extra: dict) -> dict:
+    header = {
+        "title": {"runs": [{"text": "My Playlist"}]},
+        "thumbnail": {
+            "musicThumbnailRenderer": {
+                "thumbnail": {"thumbnails": [{"url": "https://example.com/t.jpg", "width": 60, "height": 60}]}
+            }
+        },
+    }
+    header.update(extra)
+    return header
+
+
+class TestParsePlaylistHeaderMeta:
+    def test_missing_second_subtitle(self):
+        # some playlists are served without a secondSubtitle at all
+        parsed = parse_playlist_header_meta(_header())
+        assert parsed["title"] == "My Playlist"
+        assert parsed["views"] is None
+        assert parsed["duration"] is None
+        assert parsed["trackCount"] is None
+
+    def test_second_subtitle_track_count(self):
+        parsed = parse_playlist_header_meta(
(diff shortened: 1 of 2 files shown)


sigma67:
Are you sure this is still needed? do you have any examples where this still applies?

Sanjays2402:
Yes — the concrete example is issue #892: playlist `PL8eCD0llEGnywNAITfRGFiCCNpmaHc1_l` returned a header without `secondSubtitle`, and the reported traceback is exactly the unguarded access in `parse_playlist_header_meta`. The reporter later said that playlist had started returning more subtitle data, so it is intermittent rather than permanently reproducible. That is still a valid reason for the defensive guard: the response shape can omit the field, and treating the missing metadata as no views/duration/track count is safe. The PR includes a parser regression test for that shape. If you only want fixes for examples that still reproduce against the live API, I’m happy to close this instead.

sigma67:
Ok, due to

> (Or at least used to, it seems to have fixed itself now)

I thought it no longer applied, but I just checked and it still crashes

sigma67 (approved):

Issue #983, opened 2026-08-10
Song/Video playlist-item model


sigma67:
Part of #307. **Base branch: `ytmusicapi-2`.** Read the contributor guide from #982 before starting.

The single most reused shape in the library. It appears in playlist items, library songs, search results, watch playlists and history, so it gates most of Phase 2 and is worth getting right before the rest.

### Scope

Model the song/video item produced by `parsers/playlists.py::parse_playlist_items` and friends: `videoId`, `title`, `artists`, `album`, `likeStatus`, `thumbnails`, `isAvailable`, `isExplicit`, `duration`, `duration_seconds`, `setVideoId`, `videoType`, feedback tokens.

### Watch out for

- **Unavailable songs.** These are the main source of missing fields — several keys are absent where a normal song has them. This is the edge case that most needs covering.
- `duration` and `duration_seconds` travel together, cf. `parsers/_utils.py::parse_duration`.
- `videoType` and `likeStatus` should use the existing enums from `models/content/enums.py`.
- Songs and videos differ. Decide explicitly whether that is one model with optional fields or two, and write the reasoning into the PR.

### Definition of done

- Parsers construct the model; return annotations updated where the shape is now known.
- Tests cover an available song, an unavailable song and a video.
- `mypy --strict` clean.

Issue #981, opened 2026-08-10
Shared leaf models (Thumbnail, ArtistRef, AlbumRef, PlaylistRef)


sigma67:
Part of #307. **Base branch: `ytmusicapi-2`.**

Defines the small shared models that nearly every response model composes. These must exist before the core entity models (#983–#985) can be written.

### Scope

- `Thumbnail` — `{url, width, height}`
- `ArtistRef` — `{id, name}`, matching what `parsers/_utils.py::parse_id_name` produces
- `AlbumRef` — `{id, name}` album reference as embedded in songs and search results
- `PlaylistRef` — playlist reference as embedded in menus, cf. `parse_menu_playlists`
- Feedback token pair (`add`/`remove`) used by library and playlist items

Also re-home the models that already exist so everything lives under one scheme:

- `ytmusicapi/models/content/text_run.py` (`HyperLink`, `PlainText`, `TextRun`)
- `ytmusicapi/models/content/enums.py` (`PrivacyStatus`, `LikeStatus`, `VideoType`, ...)

The enums are already `str, Enum` and serialise correctly inside pydantic models — no change needed beyond placement and export.

### Definition of done

- Each model derives from `YTMusicModel` (except the enums, which stay enums).
- Exported from `ytmusicapi/models/__init__.py`.
- `mypy --strict` clean; no endpoint return type changes yet.

Issue #307, opened 2022-10-05
Type validation with pydantic


sigma67:
Something I've been meaning to implement in the long term is proper models and type validation for the responses from YouTube Music. This would provide better data validation and would notify sooner if something on the API side has changed, as opposed to the current tests which usually check for the length of the results in the response only.

[pydantic](https://github.com/pydantic/pydantic) seems to be the gold standard for this. I'd like to wait for the release of [v2.0](https://pydantic-docs.helpmanual.io/blog/pydantic-v2/) to not have to deal with breaking changes from v1 once it comes out.

Please thumbs up if you're in favor and leave a comment if you're willing to help or contribute 🥳

ambujpawar:
I am willing to help :)

sigma67:
Pydantic v2 is now in alpha: https://docs.pydantic.dev/blog/pydantic-v2-alpha/

I think we should use this issue as a starting point to create additional subtasks using GitHub tasklists: https://docs.github.com/en/issues/tracking-your-work-with-issues/about-tasklists

As this is a major endeavor with breaking changes development should take place on a v2 branch for now.

JohnHKoh:
Started playing around with this, here is a rudimentary way to provide type validation.

I created a new module called where I started to define some types. The `Playlist` type, for example, looks like this:
```
class Playlist(BaseModel):
    title: str
    playlistId: str
    thumbnails: list[Thumbnail]
    description: str
    count: Optional[str] = None
    author: Optional[list[Author]] = None
```

Then, inside "parsers/browsing.py", we can create a new instance of this `Playlist` class with the information we parse from `parse_playlist`:
```
def parse_playlist(data):
    playlist = {
        'title': nav(data, TITLE_TEXT),
        'playlistId': nav(data, TITLE + NAVIGATION_BROWSE_ID)[2:],
        'thumbnails': nav(data, THUMBNAIL_RENDERER)
    }
    subtitle = data['subtitle']
    if 'runs' in subtitle:
        playlist['description'] = "".join([run['text'] for run in subtitle['runs']])
        if len(subtitle['runs']) == 3 and re.search(r'\d+ ', nav(data, SUBTITLE2)):
            playlist['count'] = nav(data, SUBTITLE2).split(' ')[0]
            playlist['author'] = parse_song_artists_runs(subtitle['runs'][:1])

    return Playlist(**playlist) # Return playlist instance
```
[...]

PR #1008, opened 2026-09-05
feat(models): response models for the upload-library methods

### LINKED ISSUE #987: Uploads response models
Part of #307. **Base branch: `ytmusicapi-2`.** Read the contributor guide from #982 before starting.

**Good first issue** — three methods, one parser file, no dependency on the trickier core models.

### Scope

- `get_library_upload_songs`
- `get_library_upload_artist`
- `get_library_upload_album`

Models go in a new module alongside the others; parsing lives in `ytmusicapi/parsers/uploads.py`.

### Watch out for

Uploaded items are similar to but not the same as regular songs — they carry an `entityId` and lack several fields that YTM-hosted songs have. Reuse the song model from #983 only where the shape genuinely matches, otherwise define upload-specific models.

`upload_song` and `delete_upload_entity` return raw responses and are out of scope here (see #995).

### Definition of done

- All three methods return models; docstrings point at them instead of repeating JSON.
- Tests pass against the existing upload fixtures.
- `mypy --strict` clean.

atirna:
## Summary

- `get_library_upload_songs`, `get_library_upload_artist` and `get_library_upload_album` now return pydantic response models (`UploadSong`, `UploadAlbum`) deriving from the `YTMusicModel` base landed in #1004
- parsing lives in `ytmusicapi/parsers/uploads.py`: `parse_uploaded_items` constructs `UploadSong` directly, and a new `parse_upload_album` assembles the album header plus tracks
- docstrings point at the models instead of repeating the JSON shape
- `get_continuations` / `get_continuation_contents` take a TypeVar so a parse func returning models typechecks; the dict parse funcs everywhere else are unchanged

Closes #987

## Why

uploaded items are similar to but not the same as regular songs: they carry an `entityId` and lack several fields YTM-hosted songs have (`isAvailable`, `videoType`, `setVideoId`). since the song model from #983 and the shared leaf models from #981 are not merged yet, these are upload-specific models for now. the leaf shapes (`Thumbnail`, `ArtistRef`, `AlbumRef`) deliberately match what #981 plans, so re-homing them onto the shared models later is mostly a rename.

one behavioural note for 2.0 consumers: with the #307 key-presence semantics, a field the API did not return is present with value `None` instead of being absent. `"album" in song` is `True` even when the song has no album column, and `song["album"]` returns `None` instead of raising. dict-style access (`song["title"]`, `song.get("entityId")`) keeps working through the mapping protocol, so code like the existing `tests/mixins/test_uploads.py` flows is unaffected.

## Verification

- `pytest -o addopts='' tests/parsers tests/models -q`: 103 passed. new tests in `tests/parsers/test_uploads.py` cover: full song parse, missing fixedColumns / album column / thumbnails staying present as `None`, unknown likeStatus falling back to `INDIFFERENT`, menu-less items skipped, a continuation page parsed through the model parse func, and album `duration_seconds` summed over tracks (4:15 => 255, two tracks => 510)
- `mypy` (strict, repo config), clean
- `ruff check .` and `ruff format --check .`: clean
- the live tests in `tests/mixins/test_uploads.py` are unchanged and run in CI with the repo credentials

### DIFF
--- a/tests/parsers/test_uploads.py
+++ b/tests/parsers/test_uploads.py
@@ -1,0 +1,206 @@
+from ytmusicapi.continuations import get_continuations
+from ytmusicapi.models.content.enums import LikeStatus
+from ytmusicapi.models.uploads import UploadAlbum, UploadSong
+from ytmusicapi.parsers.uploads import parse_upload_album, parse_uploaded_items
+
+ARTIST_ID = "FEmusic_library_privately_owned_artist_detaila_po_CICr2crg7OWpchIIY29sZHBsYXk"
+ALBUM_ID = "FEmusic_library_privately_owned_release_detailb_po_55chars"
+
+
+def _menu(video_id: str, entity_id: str, like_status: str = "LIKE") -> dict:
+    delete_command = {
+        "confirmDialogEndpoint": {
+            "content": {
+                "confirmDialogRenderer": {
+                    "confirmButton": {
+                        "buttonRenderer": {
+                            "command": {"musicDeletePrivatelyOwnedEntityCommand": {"entityId": entity_id}}
+                        }
+                    }
+                }
+            }
+        }
+    }
+    return {
+        "menuRenderer": {
+            "items": [
+                {
+                    "menuServiceItemRenderer": {
+                        "serviceEndpoint": {"queueAddEndpoint": {"queueTarget": {"videoId": video_id}}}
+                    }
+                },
+                {"menuNavigationItemRenderer": {"navigationEndpoint": delete_command}},
+            ],
+            "topLevelButtons": [{"likeButtonRenderer": {"likeStatus": like_status}}],
+        }
+    }
+
+
+def _flex_column(run: dict) -> dict:
+    return {"musicResponsiveListItemFlexColumnRenderer": {"text": {"runs": [run]}}}
+
+
+def _linked_run(text: str, browse_id: str) -> dict:
+    return {"text": text, "navigationEndpoint": {"browseEndpoint": {"browseId": browse_id}}}
+
+
+def _upload_song_item(like_status: str = "LIKE") -> dict:
+    return {
+        "musicResponsiveListItemRenderer": {
+            "menu": _menu("Uise6RPKoek", "t_po_CICr2crg7OWpchDpjPjrBA", like_status),
+            "flexColumns": [
+                _flex_column({"text": "A Sky Full Of Stars"}),
+                _flex_column(_linked_run("Coldplay", ARTIST_ID)),
+                _flex_column(_linked_run("Ghost Stories", ALBUM_ID)),
+            ],
+            "fixedColumns": [
+                {"musicResponsiveListItemFixedColumnRenderer": {"text": {"runs": [{"text": "4:15"}]}}}
+            ],
+            "thumbnail": {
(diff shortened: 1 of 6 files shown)


ytmusicapi/models/uploads.py
```diff
@@ -0,0 +1,59 @@
...
+    width: int | None = None
+    height: int | None = None
+
+
+class ArtistRef(YTMusicModel):
+    name: str | None = None
+    id: str | None = None
+
+
+class AlbumRef(YTMusicModel):
```
sigma67: ArtistRef and AlbumRef are byte-for-byte identical classes; candidate for a shared type

tests/parsers/test_uploads.py
```diff
@@ -0,0 +1,206 @@
...
+        assert song.duration == "4:15"
+        assert song.duration_seconds == 255
+        assert song.artists[0].name == "Coldplay"
+        assert song.artists[0].id == ARTIST_ID
+        assert song.album.name == "Ghost Stories"
+        assert song.album.id == ALBUM_ID
+        assert song.likeStatus == "LIKE"
+        assert song.thumbnails[0].url == "https://i.ytimg.com/vi/Uise6RPKoek/s.jpg"
+
+    # missing data stays a present key with None value, per the #307 semantics
```
sigma67: test comment justifies behavior by citing issue #307 rather than stating the fact plainly

ytmusicapi/parsers/uploads.py
```diff
@@ -35,18 +38,26 @@ def parse_uploaded_items(results: JsonList) -> JsonList:
...
+            entityId=entityId,
+            videoId=videoId,
+            title=title,
+            duration=duration,
+            duration_seconds=parse_duration(duration),
+            artists=[ArtistRef(**artist) for artist in parse_song_artists(data, 1)],
+            album=AlbumRef(**album) if (album := parse_song_album(data, 2)) else None,
+            likeStatus=like,
+            thumbnails=thumbnails,
+        )
```
sigma67, line 51: manual ArtistRef/AlbumRef construction when pydantic already coerces nested dicts automatically (the new parse_upload_album in the same file relies on that coercion, so the two paths are inconsistent).

atirna:
thx for the review! did the first and third.

- ArtistRef/AlbumRef now share a `_NamedRef` base instead of being duplicate classes. kept both names since #981 lists them as separate shared leaf models, so this shouldnt get in the way of re-homing them later.
- fixed the test comment to just state the behaviour.

left the manual construction in place though. pydantic does coerce at runtime, but mypy rejects the named-argument form:

```
ytmusicapi/parsers/uploads.py:47: error: Argument "artists" to "UploadSong" has incompatible type "list[dict[str, Any]]"; expected "list[ArtistRef]"  [arg-type]
ytmusicapi/parsers/uploads.py:48: error: Argument "album" to "UploadSong" has incompatible type "dict[str, Any] | None"; expected "AlbumRef | None"  [arg-type]
```

`parse_upload_album` only gets away with it because `UploadAlbum(**album)` unpacks a dict, which mypy cant see through. since #981 wants `mypy --strict` clean i kept the explicit refs. happy to switch if youd rather take a cast.

mypy clean and the 8 upload parser tests pass on this head.

PR #817, opened 2025-09-07
iOS Compatibility Enhancement for ytmusicapi

Goldenfreddy0703:
## Summary
This pull request enhances ytmusicapi with comprehensive iOS compatibility across all major function categories. The changes enable seamless operation with YouTube Music's mobile single-column format while maintaining backward compatibility with desktop clients.

## Problem Statement
YouTube Music returns different response structures for mobile iOS clients compared to desktop clients:
- **Desktop**: Uses `twoColumnBrowseResultsRenderer` format
- **iOS Mobile**: Uses `singleColumnBrowseResultsRenderer` format with different navigation paths

Without iOS compatibility, many functions would fail when using mobile clients, particularly when using OAuth with the `IOS_MUSIC` client configuration.

## Solution Overview
Added iOS format detection and parsing throughout the codebase with fallback strategies:

1. **Client Configuration**: Updated default client to `IOS_MUSIC` for better OAuth compatibility
2. **Navigation Enhancement**: Enhanced navigation paths to handle both desktop and mobile formats  
3. **Parser Extensions**: Added iOS-specific parsers while maintaining existing functionality
4. **Comprehensive Coverage**: Implemented across all major function categories

### 1. Client Configuration (`ytmusicapi/helpers.py`)
```python
# Changed default client from WEB_REMIX to IOS_MUSIC
"clientName": "IOS_MUSIC",
"clientVersion": "6.42",
```

#### Library Functions (`ytmusicapi/mixins/library.py`)
- ✅ `get_library_playlists()` - iOS navigation paths + conversion
- ✅ `get_library_songs()` - Advanced iOS parsing with continuation support  
- ✅ `get_library_albums()` - iOS format detection and conversion
- ✅ `get_library_artists()` - iOS continuation handling
- ✅ `get_library_subscriptions()` - iOS-compatible navigation
- ✅ `get_library_podcasts()` - iOS format support
- ✅ `get_library_channels()` - iOS navigation enhancement
- ✅ `get_history()` - iOS path fallbacks
- ✅ `get_account_info()` - Multiple iOS navigation attempts

#### Browsing Functions (`ytmusicapi/mixins/browsing.py`)
- ✅ `get_home()` - iOS mixed content parsing
- ✅ `get_artist()` - iOS header and section parsing
- ✅ `get_artist_albums()` - iOS album format detection
- ✅ `get_user()` - iOS user content parsing
- ✅ `get_user_playlists()` - iOS playlist format conversion
[...]

PR #1009, opened 2026-09-06
fix(browsing): skip unavailable homepage cards

### LINKED ISSUE #668: Error when querying get_home()
**get_home() query no longer works**
Hello, I noticed recently that the script I used to get music suggestions no longer works. The error message isn't very clear so I'm not sure what the issue might be but the script worked a few days ago and also returns results when using something like `yt.search("Oasis Wonderwall")`...

**To Reproduce**

**Test Script  using get_home ( doesn't work)**
```py
import pprint

yt = YTMusic(oauth.json")
search_results = yt.get_home(30)
pprint.pprint(search_results)
```
**Result**
```sh
  File "/home/user/.local/share/pipx/venvs/ytmusicapi/lib/python3.12/site-packages/ytmusicapi/navigation.py", line 107, in nav
    root = root[k]
           ~~~~^^^
KeyError: 'watchEndpoint'

During handling of the above exception, another exception occurred:

  File "/home/user/.local/bin/ytmusicapi_home.py", line 14, in <module>
    search_results = yt.get_home(30)
                     ^^^^^^^^^^^^^^^
  File "/home/user/.local/share/pipx/venvs/ytmusicapi/lib/python3.12/site-packages/ytmusicapi/mixins/browsing.py", line 118, in get_home
    home.extend(parse_mixed_content(results))
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/pipx/venvs/ytmusicapi/lib/python3.12/site-packages/ytmusicapi/parsers/browsing.py", line 27, in parse_mixed_content
    content = parse_song(data)
              ^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/pipx/venvs/ytmusicapi/lib/python3.12/site-packages/ytmusicapi/parsers/browsing.py", line 86, in parse_song
    "videoId": nav(result, NAVIGATION_VIDEO_ID),
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/.local/share/pipx/venvs/ytmusicapi/lib/python3.12/site-packages/ytmusicapi/navigation.py", line 111, in nav
    raise type(e)(f"Unable to find '{k}' using path {items!r} on {root!r}, exception: {e}")
KeyError: "Unable to find 'watchEndpoint' using path ['navigationEndpoint', 'watchEndpoint', 'videoId'] on {'clickTrackingParams': 'CLECEKCzAhgAIhMIiZiNtpq6iQMVTgq3AB2ZTTys', 'browseEndpoint': {'browseEndpointContextSupportedConfigs': {'browseEndpointContextMusicConfig': {'pageType': 'MUSIC_PAGE_TYPE_ARTIST'}}}}, exception: 'watchEndpoint'"

```

Note that I just tried to generate oauth using another account and this appears to work. Also, I just modified navigation.py to print the content of the `root` variable and it does contain data, it even contains keys called `watchEndpoint` so I'm not 100% sure what the case of the issue is... could it be the some issue with the type of playlists (music Vs videos?)

Edit2: I tried using get_home(1) and it worked, then tried again and it no longer work. So perhaps an again issue with the content of the results?

Edit3: using timestamps, I was able to establish that the script does run for a few entries. However, when it lands on some it cannot find the `'watchEndpoint'` key and crashing instead of skipping to the next. Is it possible to continue rather than crashing?

bssilva06:
Closes #668.

An incomplete Listen again card like the deleted-upload example in #668 has no title browse endpoint or playback destination. `parse_mixed_content` falls through to `parse_song`, causing `get_home()` to raise `KeyError: 'watchEndpoint'` and lose all recommendations. Only parse that fallback as a song when a video ID exists; otherwise skip the card and preserve the remaining results.

The offline regression fixture is minimized from the reported response shape. It reproduces the original crash before the fix and checks that valid artist, song, and radio playlist entries retain their data and order. This does not change the shared `nav()` error handling.

Validation:
- `python -m pytest tests/parsers -o addopts= -p no:cacheprovider -q`: 61 passed. Default addopts were overridden because coverage/retry plugins were not installed.
- Ruff 0.16.0 lint and formatting checks passed for both changed files.
- mypy 2.3.0 strict project check passed for all 51 source files.
- No live-account tests were run; current live occurrence of the reported response has not been verified.

Related work: #817 adds navigation fallbacks in this parser, but still requires a browse ID absent from this example.

Prepared with AI assistance; the regression was verified locally to fail before the fix and pass afterward.


oga35767-eng:
Ty

PR #1011, opened 2026-09-11
fix get_playlist_by_id returns None as id for some playlists

Auric-Coder:


sigma67 (commented):
There is no live verification for this issue. Please provide a live playlist where this issue occurs.

